PR #1, opened 2024-05-21
Add some hints on how to get started

bwrsandman:
Wanted to try out this library to see if I could send commands to my device. Was a little confused as to where to start, so I am sharing my first steps for anyone coming after.

Notes:
* I wasn't able to get the key and token from the `cloud` module. Would be good to add a section for that
* `devices.device_selector` worked for me after getting around #2 and #3 but the attributes did not reflect the state properly and `set_temperature_target` did not change anything on the device.


rokam (approved):
Thanks

PR #14, opened 2024-05-23
build: fix container requirements

chemelli74:


chemelli74:
I see you included this change in #16 but why rejecting this PR ?
Usually is better to have a PR for each change so is easier to reviert in case of issues or for future analysis.

PR #16, opened 2024-05-24
Multiple dev improvements

rokam:
Unified #13 , #14 and task to build pkg


chemelli74 (approved):
Let's proceed with this

PR #30, opened 2024-05-27
Fix temperature out of range heat pump

rokam:
https://github.com/wuwentao/midea_ac_lan/pull/22/

### DIFF
--- a/midealocal/device.py
+++ b/midealocal/device.py
@@ -231,7 +231,7 @@ class MideaDevice(threading.Thread):
 
     def pre_process_message(self, msg: bytes) -> bool:
         if msg[9] == MessageType.query_appliance:
-            message = MessageApplianceResponse(msg)
+            message = MessageApplianceResponse(bytearray(msg))
             self._appliance_query = False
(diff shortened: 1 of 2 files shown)


midealocal/device.py
```diff
@@ -231,7 +231,7 @@ def refresh_status(self, wait_response: bool = False) -> None:
 
     def pre_process_message(self, msg: bytes) -> bool:
         if msg[9] == MessageType.query_appliance:
-            message = MessageApplianceResponse(msg)
+            message = MessageApplianceResponse(bytearray(msg))
```
chemelli74, line 234: I'm not sure if this is needed.
Especially because hen we will type the library this will create an error.

rokam: MessageApplianceResponse expects bytearray and msg is bytes. That was due to type error

chemelli74: Please git merge with latest code, `pre_process_message` is now `bytearray` ;-)

chemelli74 (approved):
Nice, thx !

wuwentao (approved):

PR #31, opened 2024-05-27
chore: fix remaining typing

chemelli74:


midealocal/security.py
```diff
@@ -80,7 +80,7 @@ def set_aes_keys(self, key: bytes | str, iv: bytes | str) -> None:
         self._aes_key = key
         self._aes_iv = iv
 
-    def aes_encrypt_with_fixed_key(self, data: str) -> bytes:
+    def aes_encrypt_with_fixed_key(self, data: bytes) -> bytes:
```
rokam, line 83: I think this change is creating the test error.

chemelli74: I don't think so, I'm 99% sure its this one:

```
  self._security.aes_decrypt_with_fixed_key(response["key"]), b"0"
```

rokam: response["key"] is str type.  Or you convert it prior calling or after.

rokam (approved):
that whould do it

PR #36, opened 2024-05-28
Unify responses in Cloud testing

rokam:
Some typing changes was needed to avoid unneeded casting


midealocal/cloud.py
```diff
@@ -178,10 +178,10 @@ async def list_appliances(
     ) -> dict[int, dict[str, Any]] | None:
         raise NotImplementedError()
 
-    async def get_device_info(self, device_id: str) -> dict[str, Any] | None:
```
chemelli74, line 181: Please check with #37 if all is fine

midealocal/cloud.py
```diff
@@ -306,7 +306,7 @@ async def list_appliances(
             return appliances
         return None
 
-    async def get_device_info(self, device_id: str) -> dict[str, Any] | None:
```
chemelli74, line 309: Same as above

midealocal/cloud.py
rokam: The issue was on test, I was unable to commit because I was passing an int to that method.

Inside the method a cast to int was made and on every usage a cast to string as made.

midealocal/cloud.py
rokam: Same as above

midealocal/cloud.py
chemelli74: Ok, didn't check myself; just asked for a double check from you ;-)

chemelli74 (approved):

PR #65, opened 2024-06-04
chore: typing e2

chemelli74:
Draft as there is a logic issue with "old_protocol"

## Summary by CodeRabbit

- **Refactor**
	- Enhanced code clarity and type safety by adding type hints to various methods and instance variables across multiple modules.

### DIFF
--- a/midealocal/message.py
+++ b/midealocal/message.py
@@ -102,7 +102,7 @@ class MessageRequest(MessageBase):
         device_type: int,
         protocol_version: int,
         message_type: int,
-        body_type: int,
+        body_type: int = NONE_VALUE,
     ) -> None:
         super().__init__()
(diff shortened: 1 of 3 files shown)


midealocal/message.py
```diff
@@ -102,7 +102,7 @@ def __init__(
         device_type: int,
         protocol_version: int,
         message_type: int,
-        body_type: int,
+        body_type: int = NONE_VALUE,
```
Necroneco, line 105: why add the default value?

Necroneco (approved):

midealocal/message.py
chemelli74: Ops, removed

Necroneco (approved):

PR #73, opened 2024-06-04
chore: typing fb

chemelli74:


midealocal/message.py
```diff
@@ -102,7 +102,7 @@ def __init__(
         device_type: int,
         protocol_version: int,
         message_type: int,
-        body_type: int,
+        body_type: int | None,
```
rokam, line 105: This doesn't look right. Body type is defined by the first byte of the body.

rokam (approved):

PR #76, opened 2024-06-04
chore: typing x13

chemelli74:


midealocal/devices/x13/message.py
```diff
@@ -59,32 +61,28 @@ def _body(self):
...
-    def __init__(self, message):
-        super().__init__(message)
+    def __init__(self, message: bytes) -> None:
+        super().__init__(bytearray(message))
         if self.body_type == 0xA4:
             self.set_body(MessageMainLightBody(super().body))
         elif self.message_type == MessageType.set and self.body_type > 0x80:
             self.set_body(MessageMainLightResponseBody(super().body))
         self.set_attr()
+        self.control_success: bool
```
rokam, line 88: Can you please explain this? I didn't get where this came from.

rokam (changes requested):

midealocal/devices/x13/message.py
chemelli74: This is a variable used but not defined in init, which is a typing issue. Also the type itself was missing

rokam (approved):

PR #79, opened 2024-06-04
chore: typing x40

chemelli74:


midealocal/devices/x40/message.py
```diff
@@ -153,12 +156,13 @@ def __init__(self, body):
...
+    def __init__(self, message: bytes) -> None:
+        super().__init__(bytearray(message))
         if (
             self.message_type
             in [MessageType.set, MessageType.notify1, MessageType.query]
             and self.body_type == 0x01
         ):
             self.set_body(Message40Body(super().body))
         self.set_attr()
+        self.fields: dict[str, Any] = {}
```
rokam, line 168: Can you please explain this? I didn't get where this came from.

rokam (changes requested):

midealocal/devices/x40/message.py
chemelli74: This is a variable used but not defined in init, which is a typing issue. Also the type itself was missing

rokam: But you are resetting its value here

chemelli74: Nope, I'm just initializing it, it's `__init__()` after all ;-)

rokam: You're right, I'm sorry

rokam (approved):

PR #85, opened 2024-06-05
chore: missing typing for AC

chemelli74:


midealocal/devices/ac/__init__.py
```diff
@@ -334,7 +336,7 @@ def set_attribute(self, attr: str, value: Any) -> None:
                     )
                     setattr(message, str(self._fresh_air_version), fresh_air)
             elif attr in self._attributes.keys():
-                message = self.make_message_uniq_set()
+                message = self.make_message_set()
```
Necroneco, line 339: Is this safe?

midealocal/devices/ac/__init__.py
```diff
@@ -354,15 +356,17 @@ def set_attribute(self, attr: str, value: Any) -> None:
...
+        )
         message.target_temperature = target_temperature
         if mode is not None:
             message.power = True
             message.mode = mode
         self.build_send(message)
 
     def set_swing(self, swing_vertical: bool, swing_horizontal: bool) -> None:
-        message: MessageGeneralSet = self.make_message_uniq_set()
+        message: MessageGeneralSet = self.make_message_set()
```
Necroneco, line 369: And this

Necroneco (changes requested):

midealocal/devices/ac/__init__.py
chemelli74: Yes, following the code `make_message_uniq_set()` it just calls 2 different methods based on a flag.
One of the 2 doesn't have the needed fields here, so...

midealocal/devices/ac/__init__.py
chemelli74: Same as above

midealocal/devices/ac/__init__.py
Necroneco: But the following code with needed fields is in `if` block...

midealocal/devices/ac/__init__.py
Necroneco: Here can assume ok

midealocal/devices/ac/__init__.py
chemelli74: You need a `MessageGeneralSet` object or it won't have the needed fields.
And the original method returns `MessageSubProtocolSet | MessageGeneralSet` based on the if.

So is safe to use always the method that returns  `MessageGeneralSet`

Necroneco: If not enter the following `if`,  `MessageSubProtocolSet` seems also ok?

chemelli74: Nope.
Check the following fields:

image

those are available only for `MessageGeneralSet` ;-)

Necroneco: Yes, I mean, if `self._used_subprotocol` is `True` and `attr` not in these attributes,  is this possible?

Necroneco: I didn't go too deep into the code, just asked questions about this control flow.
If this situation is impossible, then there is no problem.

Necroneco: Ok, I can give approval, but I still think this could be a change

Necroneco: https://github.com/rokam/midea-local/pull/33#issuecomment-2254546944

PR #88, opened 2024-06-05
build: pre-commit hook

chemelli74:
## Summary by CodeRabbit

- **Chores**
  - Added a pre-commit step in the build workflow to enhance code quality checks.

- **Bug Fixes**
  - Improved type handling in the device module to ensure accurate protocol determination.

### DIFF
--- a/midealocal/devices/e2/__init__.py
+++ b/midealocal/devices/e2/__init__.py
@@ -1,7 +1,7 @@
 import json
 import logging
 import sys
-from typing import Any, cast
+from typing import Any
 
 from .message import (
     MessageE2Response,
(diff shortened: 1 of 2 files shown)


midealocal/devices/e2/__init__.py
```diff
@@ -90,13 +90,17 @@ def _normalize_old_protocol(self, value: Any) -> OldProtocol:
...
+                        or self.subtype == 85
+                        or self.subtype == 36353
+                    )
                     value = OldProtocol.true if result else OldProtocol.false
             elif isinstance(value, int) or isinstance(value, bool):
                 value = OldProtocol.true if value else OldProtocol.false
             else:
                 raise ValueError("Invalid value for old_protocol")
-            return value
+            return cast(OldProtocol, value)
```
rokam, line 103: This can hide mistakes. I do prefer if you use a different variable to return. Maybe called return_value instead of value.

rokam (changes requested):

midealocal/devices/e2/__init__.py
chemelli74: Indeed, you are right. Updated

rokam (approved):

PR #93, opened 2024-06-05
fix: normalize old_protocol for E2

chemelli74:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new `OldProtocol` enum with values `auto`, `true`, and `false`.
  - Enhanced the logic for handling old protocol settings.

- **Refactor**
  - Improved the `old_protocol` method, now `_normalize_old_protocol`, for better performance and reliability.
  - Updated the `set_customize` method to more robustly handle the `old_protocol` parameter.

### DIFF
--- a/midealocal/devices/e2/__init__.py
+++ b/midealocal/devices/e2/__init__.py
@@ -18,6 +18,13 @@ else:
 
 from ...device import MideaDevice
 
+
+class OldProtocol(StrEnum):
+    auto = "auto"
+    true = "true"
+    false = "false"
+
+
 _LOGGER = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


midealocal/devices/e2/__init__.py
```diff
@@ -74,12 +81,22 @@ def __init__(
...
         self._old_protocol = self._default_old_protocol
         self.set_customize(customize)
 
-    def old_protocol(self) -> bool:
-        return self.subtype <= 82 or self.subtype == 85 or self.subtype == 36353
+    def _normalize_old_protocol(self, value: Any) -> OldProtocol:
+        try:
+            if isinstance(value, str):
+                value = OldProtocol(value)
+            elif isinstance(value, int):
```
rokam, line 92:
```suggestion
            elif isinstance(value, int) or isinstance(value, bool):

```

rokam (changes requested):
We should consider it bool too

midealocal/devices/e2/__init__.py
```diff
@@ -74,12 +81,22 @@ def __init__(
...
+                value = OldProtocol.true if value else OldProtocol.false
+            if not isinstance(value, OldProtocol):
```
rokam, line 94: Better remove unneeded comparison
```suggestion
            else:
```

rokam (dismissed):

chemelli74:
> We should consider it bool too

Done, even if I think that once the user enter False in the dialog is will still be saved as string.

Necroneco (approved):

Necroneco:
> Done, even if I think that once the user enter False in the dialog is will still be saved as string.

The user input is a complete JSON string, it will be parsed as expected.

rokam (approved):

PR #97, opened 2024-06-06
change the string of fan speed to be consistent with the constant from HA

Necroneco:
wuwentao/midea_ac_lan#72

Only changed `AC` and `CC`.

No changes to `CC` devices with 7 levels

---

The modification seems safe, but I'm not very sure.

### DIFF
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -64,12 +64,12 @@ class DeviceAttributes(StrEnum):
 
 class MideaACDevice(MideaDevice):
     _fresh_air_fan_speeds = {
-        0: "Off",
-        20: "Silent",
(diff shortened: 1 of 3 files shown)


rokam:
This change could cause issues. We need to map it to a major version bump.

chemelli74:
> No changes to `CC` devices with 7 levels

Why ?

Necroneco:
> Why ?

I don't know how to modify it, simply change it to lower case? Or change some of the levels to low, medium, high?

chemelli74:
> > Why ?
> 
> I don't know how to modify it, simply change it to lower case? Or change some of the levels to low, medium, high?

I would use the same approach as for others at least for now.
If and when we'll have a  issue report then we can investigate further.

PR #127, opened 2024-06-14
chore: initial fix for ruff ANN401

chemelli74:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new exception class `DataSignWrongType` to handle incorrect data types during signing.

- **Improvements**
  - Enhanced type safety for various methods and functions, improving code reliability and maintainability.
  - Updated method signatures and return types to provide clearer and more specific type expectations.

These improvements ensure better error handling and more predictable behavior within the application.

### DIFF
--- a/midealocal/device.py
+++ b/midealocal/device.py
@@ -543,7 +543,8 @@ class MideaDevice(threading.Thread):
         raise NotImplementedError
 
     def get_attribute(self, attr: str) -> bool | int | str | None:
-    """Get attribute."""
+        """Get attribute."""
+
         return self._attributes.get(attr)
 
     def set_customize(self, customize: str) -> None:
--- a/midealocal/device.py
+++ b/midealocal/device.py
@@ -543,6 +543,7 @@ class MideaDevice(threading.Thread):
         raise NotImplementedError
 
     def get_attribute(self, attr: str) -> bool | int | str | None:
+    """Get attribute."""
         return self._attributes.get(attr)
 
     def set_customize(self, customize: str) -> None:
--- a/midealocal/device.py
+++ b/midealocal/device.py
@@ -542,8 +542,7 @@ class MideaDevice(threading.Thread):
         """Set attribute."""
         raise NotImplementedError
 
-    def get_attribute(self, attr: str) -> Any:
-        """Get attribute."""
+    def get_attribute(self, attr: str) -> bool | int | str | None:
         return self._attributes.get(attr)
 
     def set_customize(self, customize: str) -> None:
--- a/midealocal/devices/__init__.py
+++ b/midealocal/devices/__init__.py
@@ -1,9 +1,119 @@
 """Midea local devices."""
 
 from importlib import import_module
-from typing import Any
+from typing import cast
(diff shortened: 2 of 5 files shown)


midealocal/security.py
```diff
@@ -212,7 +212,9 @@ class MideaAirSecurity(CloudSecurity):
     def __init__(self, login_key: str) -> None:
         super().__init__(login_key, None, None)
 
-    def sign(self, url: str, data: Any, random: str) -> str:
+    def sign(self, url: str, data: dict[str, Any] | str, random: str) -> str:
+        if isinstance(data, str):
```
rokam, line 216: This should be outside of this PR

midealocal/devices/__init__.py
```diff
@@ -126,7 +127,7 @@ def device_selector(
     model: str,
     subtype: int,
     customize: str,
-) -> Any:
+) -> MideaDevice:
```
rokam, line 130: Maybe we should deprecate this method. I don't like this dynamic importing. What is you thought about this?

rokam (changes requested):

midealocal/devices/__init__.py
chemelli74: I agree, would you please take care of it ?

midealocal/security.py
chemelli74: SO you prefer a single PR with the check and after this typing improvement ?

midealocal/security.py
```diff
@@ -218,7 +218,9 @@ class MideaAirSecurity(CloudSecurity):
     def __init__(self, login_key: str) -> None:
         super().__init__(login_key, None, None)
 
-    def sign(self, url: str, data: Any, random: str) -> str:
+    def sign(self, url: str, data: dict[str, Any] | str, random: str) -> str:
+        if isinstance(data, str):
+            raise Exception("wrong data type to sign")
```
rokam, line 223: Need to change to a specific exception.

rokam (changes requested):

midealocal/device.py
```diff
@@ -542,8 +542,7 @@ def set_attribute(self, attr: str, value: Any) -> None:
         """Set attribute."""
         raise NotImplementedError
 
-    def get_attribute(self, attr: str) -> Any:
-        """Get attribute."""
+    def get_attribute(self, attr: str) -> bool | int | str | None:
```
rokam, line 545:
```suggestion
    def get_attribute(self, attr: str) -> bool | int | str | None:
    """Get attribute."""
```

midealocal/device.py
```diff
@@ -542,8 +542,8 @@ def set_attribute(self, attr: str, value: Any) -> None:
         """Set attribute."""
         raise NotImplementedError
 
-    def get_attribute(self, attr: str) -> Any:
-        """Get attribute."""
+    def get_attribute(self, attr: str) -> bool | int | str | None:
+    """Get attribute."""
```
rokam, line 546:
```suggestion
        """Get attribute."""

```

rokam (approved):

PR #128, opened 2024-06-14
feat: expose cloud servers

chemelli74:
Expose cloud servers list so Home Assistant components can grab them easily and avoid duplicating data.

## Summary by CodeRabbit

- **New Features**
  - Introduced functionality to retrieve available cloud servers, enhancing the app's cloud management capabilities.

### DIFF
--- a/midealocal/cloud.py
+++ b/midealocal/cloud.py
@@ -192,7 +192,7 @@ class MideaCloud:
         return result
 
     async def get_cloud_servers(self) -> dict[int, str]:
-        """Get available cloud servers from upstream library."""
+        """Get available cloud servers."""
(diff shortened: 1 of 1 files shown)


midealocal/cloud.py
```diff
@@ -191,6 +191,13 @@ async def get_keys(self, appliance_id: int) -> dict[int, dict[str, Any]]:
         result.update(default_keys)
         return result
 
+    async def get_cloud_servers(self) -> dict[int, str]:
```
rokam, line 194: This could be static,
```suggestion
    @staticmethod
    async def get_cloud_servers(self) -> dict[int, str]:
```

rokam (approved):

PR #161, opened 2024-06-16
chore: ruff ANN401

rokam:
## Summary by CodeRabbit

- **New Features**
  - Enhanced type safety across various device modules by updating the `set_attribute` method to accept specific types (`int`, `str`, `bool`).
  - Added a new `ValueWrongType` exception for handling incorrect data type errors.

- **Bug Fixes**
  - Improved attribute value handling in Midea A1, AC, B3, B4, x26, x34, and x40 device modules to ensure correct type conversions and settings.

- **Tests**
  - Updated `test_set_attribute` in `device_ac_test.py` to reflect changes in method calls and type validation.

- **Refactor**
  - Added type hints and annotations to various methods and attributes for better code clarity and maintainability.

### DIFF
--- a/midealocal/device.py
+++ b/midealocal/device.py
@@ -5,7 +5,7 @@ import socket
 import threading
 import time
 from enum import IntEnum, StrEnum
-from typing import Any
+from typing import Any, Callable
 
 from .exceptions import SocketException
 from .message import (
@@ -136,7 +136,7 @@ class MideaDevice(threading.Thread):
         self._model = model
         self._subtype = subtype
         self._protocol_version = 0
-        self._updates = [dict]
+        self._updates: list[Callable[[dict[str, Any]], None]] = []
         self._unsupported_protocol = [dict]
         self._is_run = False
         self._available = True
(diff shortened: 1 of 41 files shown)


midealocal/device.py
```diff
@@ -554,7 +554,7 @@ def run(self) -> None:
                     self.close_socket()
                     break
 
-    def set_attribute(self, attr: str, value: Any) -> None:
+    def set_attribute(self, attr: str, value: Any) -> None:  # noqa: ANN401
```
chemelli74, line 557:
```suggestion
    def set_attribute(self, attr: str, value: bool | int | str) -> None: 
```

midealocal/devices/b0/__init__.py
```diff
@@ -100,7 +100,7 @@ def process_message(self, msg: bytes) -> dict:
                 new_status[str(status)] = self._attributes[status]
         return new_status
 
-    def set_attribute(self, attr: str, value: Any) -> None:
+    def set_attribute(self, attr: str, value: Any) -> None:  # noqa: ANN401
```
chemelli74, line 103:
```suggestion
    def set_attribute(self, attr: str, value: bool | int | str) -> None:
```

chemelli74 (changes requested):
I see `set_attribute` has different signatures. Can we fix them ?

rokam:
> I see `set_attribute` has different signatures. Can we fix them ?

I don't know if this is better then leaving the right scope in each device. That's why I left some with Any.

chemelli74 (approved):
Very nice one, thx !

PR #177, opened 2024-06-18
fix: allign set_target_temperature signatures (part2)

chemelli74:
## Summary by CodeRabbit

- **New Features**
	- Introduced the ability to set target temperature for Midea FB devices with optional mode and zone parameters.

- **Improvements**
	- Enhanced Midea C3 device temperature setting functionality by adding a default value for the zone parameter.

### DIFF
--- a/midealocal/devices/c3/__init__.py
+++ b/midealocal/devices/c3/__init__.py
@@ -290,11 +290,11 @@ class MideaC3Device(MideaDevice):
         self,
         target_temperature: float,
         mode: int | None,
-        zone: int,
+        zone: int | None = None,
     ) -> None:
         """Midea C3 device set target temperature."""
         message = self.make_message_set()
-        if self._attributes[DeviceAttributes.zone_temp_type][zone]:
+        if zone and self._attributes[DeviceAttributes.zone_temp_type][zone]:
             message.zone_target_temp[zone] = target_temperature
         else:
             message.room_target_temp = target_temperature
--- a/midealocal/devices/fb/__init__.py
+++ b/midealocal/devices/fb/__init__.py
@@ -112,6 +112,14 @@ class MideaFBDevice(MideaDevice):
             setattr(message, str(attr), value)
         self.build_send(message)
 
+    def set_target_temperature(
+        self,
+        target_temperature: float,
+        mode: int | None,
+        zone: int | None = None,
+    ) -> None:
+        """Midea FB device set target temperature."""
+
 
 class MideaAppliance(MideaFBDevice):
     """Midea FB appliance."""


rokam (approved):

midealocal/devices/c3/__init__.py
```diff
@@ -290,11 +290,11 @@ def set_target_temperature(
         self,
         target_temperature: float,
         mode: int | None,
-        zone: int,
+        zone: int | None = None,
     ) -> None:
         """Midea C3 device set target temperature."""
         message = self.make_message_set()
-        if self._attributes[DeviceAttributes.zone_temp_type][zone]:
+        if zone and self._attributes[DeviceAttributes.zone_temp_type][zone]:
```
Necroneco, line 297: What happens if zone == 0?

midealocal/devices/fb/__init__.py
```diff
@@ -112,6 +112,14 @@ def set_attribute(self, attr: str, value: str | int | bool) -> None:
             setattr(message, str(attr), value)
         self.build_send(message)
 
+    def set_target_temperature(
```
Necroneco, line 115: Why add `set_target_temperature` to FB device?

chemelli74: BEacuse is needed on integration side as for this device we use directly set_attribute which has the wrong params

midealocal/devices/c3/__init__.py
chemelli74: is 0 a valid value ? I think zone is >=1 or None

Necroneco: Obviously `zone` is 0 or 1

```python
            message.zone_target_temp[zone] = target_temperature
```

```python
            if zone == 0:
                message.zone1_power = True
            else:
                message.zone2_power = True
```

```python
        zone1_target_temp = int(self.zone_target_temp[0])
        zone2_target_temp = int(self.zone_target_temp[1])
```

chemelli74: > Obviously `zone` is 0 or 1

or None ;-)

chemelli74: Is enough to change the line to "if zone is not None..."

chemelli74: #180

midealocal/devices/fb/__init__.py
chemelli74: #179

midealocal/devices/c3/__init__.py
Necroneco: `None` is not valid value, maybe should raise an error

PR #180, opened 2024-06-19
fix(c3): zone check

chemelli74:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability of temperature settings by ensuring proper handling of zone-specific configurations.

### DIFF
--- a/midealocal/devices/c3/__init__.py
+++ b/midealocal/devices/c3/__init__.py
@@ -293,8 +293,11 @@ class MideaC3Device(MideaDevice):
         zone: int | None = None,
     ) -> None:
         """Midea C3 device set target temperature."""
+        if zone is None:
+            raise ValueError("[C3] Parameter `zone` must be set")
+
         message = self.make_message_set()
-        if zone is not None and self._attributes[DeviceAttributes.zone_temp_type][zone]:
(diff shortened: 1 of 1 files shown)


Necroneco:
`if zone is not None` and `else`, make `None` looks like a valid value, but it's not.
I think raise an error maybe better

rokam (approved):

Necroneco (approved):

midealocal/devices/c3/__init__.py
```diff
@@ -297,7 +297,7 @@ def set_target_temperature(
             raise ValueError("[C3] Parameter `zone` must be set")
 
         message = self.make_message_set()
-        if self._attributes[DeviceAttributes.zone_temp_type][zone]:
+        if zone and self._attributes[DeviceAttributes.zone_temp_type][zone]:
```
Necroneco, line 300: why restore this

Necroneco (changes requested):

midealocal/devices/c3/__init__.py
chemelli74: thx, fixed

Necroneco (approved):

PR #194, opened 2024-07-01
chore(library_test): more details on login failure

chemelli74:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error messages for cloud service login failures by including device details, offering better context for users.

- **Chores**
  - Removed unnecessary linting rule for `print` statements to streamline development.

### DIFF
--- a/library_test.py
+++ b/library_test.py
@@ -84,6 +84,7 @@ async def main() -> None:
             msg = f"\nERROR: cannot login into device {first_device['device_id']} \
 [{first_device['ip_address']}]\n"
             print(msg)
+            await session.close()
             sys.exit(2)
         cloud_keys = await cloud.get_keys(first_device["device_id"])
     print("-" * 20)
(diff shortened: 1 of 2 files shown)


rokam (approved):

rokam (changes requested):
It's better to use the color log.

wuwentao (approved):

rokam:
> > It's better to use the color log.
> 
> So you want to rework the full library to use logger and not print, right ?

Just the CLI. You've added color log to the library_test.py so an error here should be shown as an Error.

chemelli74:
> > > It's better to use the color log.
> > 
> > 
> > So you want to rework the full library to use logger and not print, right ?
> 
> Just the CLI. You've added color log to the library_test.py so an error here should be shown as an Error.

Yes of course, I meant the `library_test` ;-)
In other project I used colorlog for lines coming from the library itself, and print (white) for message from the test script
From my poinf of view this helps understand where the message comes from.

But if you want me to change all print, I'll do.

rokam (approved):

PR #199, opened 2024-07-03
fix(da): avoid indexes out of bounds

rokam:
## Summary by CodeRabbit

- **Bug Fixes**
  - Updated logic for assigning device attributes to ensure correct handling based on attribute values and array lengths.

- **Tests**
  - Added comprehensive test cases for initial attribute settings, message processing, query building, and attribute setting functionalities of the `MideaDADevice` class.
  - Added test cases for various message types related to the DA system, including query, power, start, and response messages.

### DIFF
--- a/midealocal/devices/da/__init__.py
+++ b/midealocal/devices/da/__init__.py
@@ -140,13 +140,13 @@ class MideaDADevice(MideaDevice):
                 if status == DeviceAttributes.progress:
                     self._attributes[status] = (
                         "-"
-                        if status >= len(progress)
+                        if getattr(message, str(status)) >= len(progress)
                         else progress[getattr(message, str(status))]
                     )
                 elif status == DeviceAttributes.program:
                     self._attributes[status] = (
                         "-"
-                        if status >= len(program)
+                        if getattr(message, str(status)) >= len(program)
                         else program[getattr(message, str(status))]
                     )
                 elif status == DeviceAttributes.rinse_level:
@@ -158,25 +158,25 @@ class MideaDADevice(MideaDevice):
                 elif status == DeviceAttributes.dehydration_speed:
                     self._attributes[status] = (
                         "-"
-                        if status >= len(speed)
+                        if getattr(message, str(status)) >= len(speed)
                         else speed[getattr(message, str(status))]
                     )
                 elif status == DeviceAttributes.detergent:
                     self._attributes[status] = (
                         "-"
-                        if status >= len(detergent)
+                        if getattr(message, str(status)) >= len(detergent)
                         else detergent[getattr(message, str(status))]
                     )
                 elif status == DeviceAttributes.softener:
                     self._attributes[status] = (
                         "-"
-                        if status >= len(softener)
+                        if getattr(message, str(status)) >= len(softener)
                         else softener[getattr(message, str(status))]
                     )
                 elif status == DeviceAttributes.wash_strength:
                     self._attributes[status] = (
                         "-"
-                        if status >= len(strength)
+                        if getattr(message, str(status)) >= len(strength)
                         else strength[getattr(message, str(status))]
                     )
                 else:
--- a/midealocal/devices/da/__init__.py
+++ b/midealocal/devices/da/__init__.py
@@ -147,11 +147,7 @@ class MideaDADevice(MideaDevice):
                         None if value >= len(program) else program[value]
                     )
                 elif status == DeviceAttributes.rinse_level:
-                    temp_rinse_level = value
-                    if temp_rinse_level == MIN_TEMP:
-                        self._attributes[status] = None
-                    else:
-                        self._attributes[status] = temp_rinse_level
+                    self._attributes[status] = "-" if value == MIN_TEMP else value
                 elif status == DeviceAttributes.dehydration_speed:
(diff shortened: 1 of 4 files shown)


midealocal/devices/da/__init__.py
```diff
@@ -140,13 +140,13 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
                 if status == DeviceAttributes.progress:
                     self._attributes[status] = (
                         "-"
-                        if status >= len(progress)
+                        if getattr(message, str(status)) >= len(progress)
```
chemelli74, line 143: Please save the result in a variable before the if and reuse it in both options

chemelli74 (changes requested):
Added a comment, that is valid for all similar block places

midealocal/devices/da/__init__.py
```diff
@@ -137,30 +137,39 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
                     else:
                         self._attributes[status] = temp_rinse_level
                 elif status == DeviceAttributes.dehydration_speed:
-                    temp_speed = getattr(message, str(status))
-                    if temp_speed == MIN_TEMP:
-                        self._attributes[status] = "-"
-                    else:
-                        self._attributes[status] = speed[temp_speed]
+                    self._attributes[status] = (
+                        "-" if value >= len(speed) else speed[value]
```
chemelli74, line 157: None ?

midealocal/devices/da/__init__.py
```diff
@@ -137,30 +137,39 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+                    )
                 elif status == DeviceAttributes.softener:
-                    self._attributes[status] = softener[getattr(message, str(status))]
+                    self._attributes[status] = (
+                        None if value >= len(softener) else softener[value]
+                    )
                 elif status == DeviceAttributes.wash_strength:
-                    self._attributes[status] = strength[getattr(message, str(status))]
+                    self._attributes[status] = (
+                        "-" if value >= len(strength) else strength[value]
```
chemelli74, line 169: None ?

chemelli74 (changes requested):

midealocal/devices/da/__init__.py
rokam: The first item of the array is "-". Thats why I left - instead of None.

midealocal/devices/da/__init__.py
rokam: The first item of the array is "-". Thats why I left - instead of None.

midealocal/devices/da/__init__.py
```diff
@@ -137,30 +137,39 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+                        None if value >= len(progress) else progress[value]
+                    )
                 elif status == DeviceAttributes.program:
-                    self._attributes[status] = program[getattr(message, str(status))]
+                    self._attributes[status] = (
+                        None if value >= len(program) else program[value]
+                    )
                 elif status == DeviceAttributes.rinse_level:
-                    temp_rinse_level = getattr(message, str(status))
+                    temp_rinse_level = value
```
chemelli74, line 150: can we remove temp_rinse_level e use directly value ?

midealocal/devices/da/__init__.py
```diff
@@ -137,30 +137,39 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
                         self._attributes[status] = temp_rinse_level
                 elif status == DeviceAttributes.dehydration_speed:
-                    temp_speed = getattr(message, str(status))
-                    if temp_speed == MIN_TEMP:
-                        self._attributes[status] = "-"
-                    else:
-                        self._attributes[status] = speed[temp_speed]
+                    self._attributes[status] = (
+                        None if value >= len(speed) else speed[value]
+                    )
```
chemelli74, line 158: Seems here we slightly changed logic for MIN_TEMP

rokam: Should we return None or "-"?

chemelli74: I already voted for "None", but if I get it right, now you return None only if value >= len(speed) and not also when equal to MIN_TEMP. Am I wrong ?

midealocal/devices/da/__init__.py
```diff
@@ -137,30 +137,35 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+                    self._attributes[status] = (
+                        None if value >= len(program) else program[value]
+                    )
                 elif status == DeviceAttributes.rinse_level:
-                    temp_rinse_level = getattr(message, str(status))
-                    if temp_rinse_level == MIN_TEMP:
-                        self._attributes[status] = "-"
-                    else:
-                        self._attributes[status] = temp_rinse_level
+                    self._attributes[status] = "-" if value == MIN_TEMP else value
```
chemelli74, line 150: Didn't we voted for `None` ?

rokam: I did just return to "-" for rinse_level, to keep the current behaviour.

chemelli74 (approved):
Good for now, but we should rework to completely remove "-"

PR #200, opened 2024-07-04
fix: rollback get_keys with 3 keys

wuwentao:
in current changes, get_keys will only return method 1 token/key and skip method 2 and default key.

in previous design, we should return 3 token/keys, and try with all the possible token/key.

use more token/key to retry always better than only check with one key.

so we need to rollback it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Ensured the `get_keys` functionality always updates with default keys, improving reliability of the feature.

- **Tests**
  - Expanded test coverage to include additional scenarios for `get_keys`, ensuring better validation and reliability.

- **Chores**
  - Updated JSON response files to align with new key naming conventions.

### DIFF
--- a/midealocal/cloud.py
+++ b/midealocal/cloud.py
@@ -206,9 +206,8 @@ class MideaCloud:
                             "token": token["token"].lower(),
                             "key": token["key"].lower(),
                         }
-                break
-        if not result:
-            result.update(default_keys)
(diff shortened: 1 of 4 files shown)


rokam:
You need to adjust tests.

rokam (approved):

PR #201, opened 2024-07-04
chore: check PR conventional commit name

rokam:
## Summary by CodeRabbit

- **Chores**
  - Introduced a GitHub Actions workflow to validate pull request titles.
  - Added a commit linting job to the existing Python build workflow.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,12 @@
+name: PR
+on: pull_request
+jobs:
+  conventional:
+    name: Conventional PR
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+      - uses: beemojs/conventional-pr-action@v3
+        env:
+          GITHUB_TOKEN: ${{ secrets.MIDEA_GITHUB_PAT }}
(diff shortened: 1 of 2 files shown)


chemelli74:
I would suggest going with commlint.
Ex: https://github.com/chemelli74/aiocomelit/blob/main/.github/workflows/ci.yml#L27-L36

rokam:
> I would suggest going with commlint. Ex: https://github.com/chemelli74/aiocomelit/blob/main/.github/workflows/ci.yml#L27-L36

Sure is useful, but I wan´t to check if the PR name is correct, commitlint checks the commits inside the PR.

chemelli74:
> > I would suggest going with commlint. Ex: https://github.com/chemelli74/aiocomelit/blob/main/.github/workflows/ci.yml#L27-L36
> 
> Sure is useful, but I wan´t to check if the PR name is correct, commitlint checks the commits inside the PR. And, if we use squash, those commits are irrelevant.

Not exactly, commlint check that the PR commit is compliant to a schema. Like you are trying to achive with another hook.
If you name the PR wrongly, it won't pass. You can easily make a test if you prefer to double check.

rokam:
I did both, one action just for PR name and a lint in the ci job.

chemelli74 (approved):

PR #209, opened 2024-07-08
feat(c3): temperature step as customize

rokam:
## Summary by CodeRabbit

- **New Features**
  - Added ability to set a custom temperature step for Midea C3 devices.

- **Tests**
  - Introduced tests for the Midea C3 device to ensure proper functionality and handling of various scenarios.

### DIFF
--- a/midealocal/devices/c3/__init__.py
+++ b/midealocal/devices/c3/__init__.py
@@ -1,5 +1,6 @@
 """Midea local C3 device."""
 
+import json
 import logging
 from enum import IntEnum, StrEnum
 from typing import Any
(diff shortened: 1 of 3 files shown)


midealocal/devices/c3/__init__.py
```diff
@@ -309,6 +318,29 @@ def set_target_temperature(
             message.mode = mode
         self.build_send(message)
 
+    def set_customize(self, customize: str) -> None:
```
chemelli74, line 321: Can we place the `set_customize` in a common class and use for all devices ?

rokam: I don't like this. Every device type has its own logic for customize. If we'll refact that, it should go in a different PR.

chemelli74: I think we are making a bit of spaghetti code, we have som many common places copied&pasted all over.

I'll approve this one, but we should really clean up the code before adding more caos.

PR #231, opened 2024-07-18
feat: parse all the response items from MeijuCloud for get_device_inf

wuwentao:
## Summary by CodeRabbit

- **New Features**
  - Enhanced device information with additional attributes like description, active status, active time, master ID, WiFi version, and more, improving data handling capabilities.

- **Documentation**
  - Updated docstrings to provide clearer information on API interaction, including details on expected input and response structure.

- **Tests**
  - Added multiple assertions in tests to validate new device attributes, enhancing test coverage and ensuring reliable device information retrieval.


rokam (changes requested):
Please improve tests to check the parse process

midealocal/cloud.py
```diff
@@ -408,7 +408,25 @@ async def list_appliances(
         return None
 
     async def get_device_info(self, device_id: int) -> dict[str, Any] | None:
-        """Get device information."""
+        r"""Get device information.
```
chemelli74, line 411:
```suggestion
        """Get device information.
```

chemelli74 (changes requested):

midealocal/cloud.py
wuwentao: oh, thanks for your review comments, seems without `r`, it can't accept multiple lines by pre-commit, so I just simple add `r` to fix it, is there any good solution to prevent pre-commit error with this DOC str in multiple lines?
the error from vscode is 
```
Use `r"""`  if any backslashes in a docstring  
```
from Ruff(D301).

chemelli74: I would avoid reporting an example of each value, while writing the type.
An example below.

midealocal/cloud.py
```diff
@@ -408,7 +408,25 @@ async def list_appliances(
...
+
+        API url: https://mp-prod.smartmidea.net/mas/v5/app/proxy?alias=/v1/appliance/info/get
+        header:
+        input: {'applianceCode': 21000***830**18,
+            'reqId': 'b11bb9083be6d77906fe1c9f019cdea0', 'stamp': '20240710092728'}
+        response: b'{"code":0,"msg":null,"data":{"id":null,
+            "applianceCode":21000***830**18,
+            "sn":"7105f17f36a6afcce272f8053e2be60fd74b1a4baca120afaad83011bb50e8d5f3678bf88e32ea11885394e1a32c9c0e",
+            "onlineStatus":1,"type":"0xDB","modelNumber":"12877",
+            "name":"\xe6\xbb\x9a\xe7\xad\x92\xe6\xb4\x97\xe8\xa1\xa3\xe6\x9c\xba",
```
chemelli74, line 421:
```suggestion
            "onlineStatus":int,
            "type": byte,
            "modelNumber": string,
            "name": bytearray,
```

chemelli74 (changes requested):

midealocal/cloud.py
wuwentao: it's document info, and only used for our future reference when we change source code, without the origin example, we can't know why we add it, the type is origin from cloud response, not from our code, it's not trust and maybe change or maybe difference based on difference cloud service, so i don't think it's a good idea to remove the origin msg if different cloud return different type or result. we should have a normal reference, for now and also for future.

wuwentao:
> Please improve tests to check the parse process

thanks for your review comments, add test done, please help to review and check the testcase info.

midealocal/cloud.py
chemelli74: Example documentation should not be in the code. We must only document that we expect a "int" or a "string", log the message at debug level.
In future, it things changes you have those references and stop the change easily.

chemelli74 (dismissed):

midealocal/cloud.py
wuwentao: I really can understand what you expected in a normal published document or API func part, and I also did it in some project, but I really don't agree with you about this changes in current project, as it just waste time for us now.

most of us don't add any code comments or document in current source code,  and we also not publish any document with these source code doc part, just for a reference, or for future code changes, it's enough for us and also enough for current project requirement.

And with this additional original example msg in document part, we can quickly found it has changed or NOT changed, also can understanding the types.

if you still think I need to change it and add all the types, ok, I can agree with you and do these changes, but we also should follow it to add more document info with the same requirement in future.

Thanks, waiting for your response and wish we can got a agreenment.

midealocal/cloud.py
rokam, line 411:
```suggestion
        """
        Get device information.
```
This will solve it.

midealocal/cloud.py
rokam: We could add docs inside some markdown files, that way we have documentation about those apis and also we need to add docs about the device protocol, maybe the lua files should be enough.
For B8 device, I did add some comments about the protocol, but I don't think that is the best way also.

wuwentao: most of the lint tools force checking the doc str, they expect we should have a brief or detail description about it. so why we need to add it ?  and why these lint tools force check it ?

per my understand, add docs info should have some purpose: 
1.  func brief/detail description, input args, return value, etc,
2.  documentation generate with these doc str.

**it just based on our requirements, this is the most important part.**
for the func input args/return types, we already have mypy to do these works.
what we need in func doc part (or file doc part) is only the func(file) detail or description or some reference info...


in addition,  add code comments is used to quickly understand what/why it did(I don't want to remember it, just add comments for future reference, all of us may forgot it)

midealocal/cloud.py
wuwentao: done

rokam (approved):

rokam (approved):
LGTM

wuwentao:
@chemelli74 @rokam 
I'm sorry for the delay these days, any comments with it?  especially for @chemelli74 , we should have a agreement with it, for this PR and also for fututure PRs.

in addition, these response info is json format from cloud server, so all the types should be json types, and NOT python types. for example, json null vs python none.

thanks.

rokam:
> @chemelli74 @rokam 
> I'm sorry for the delay these days, any comments with it?  especially for @chemelli74 , we should have a agreement with it, for this PR and also for fututure PRs.
> 
> in addition, these response info is json format from cloud server, so all the types should be json types, and NOT python types. for example, json null vs python none.
> 
> thanks.
> 

It's ok for me.

chemelli74:
If you guys both think that's fine, go ahead

wuwentao:
thanks.

PR #233, opened 2024-07-19
feat: segregate connect/auth/refresh/enable device duties

chemelli74:
## Summary by CodeRabbit

- **New Features**
	- Enhanced device connection reliability with a retry mechanism for socket connections.
	- Introduced a new method for streamlined authentication and status refresh.
	- Improved logging for device state changes and authentication processes.
	- Added a new exception class for better error handling related to connection failures.

- **Bug Fixes**
	- Addressed potential transient connection failures through retry logic.

- **Tests**
	- Refactored tests for improved clarity and maintainability using parameterization.

### DIFF
--- a/midealocal/device.py
+++ b/midealocal/device.py
@@ -8,7 +8,7 @@ from collections.abc import Callable
 from enum import IntEnum, StrEnum
 from typing import Any
 
-from .exceptions import SocketException
+from .exceptions import CannotConnect, SocketException
 from .message import (
     MessageApplianceResponse,
     MessageQueryAppliance,
@@ -190,13 +190,11 @@ class MideaDevice(threading.Thread):
                 break
         return result, msg
 
-    def _authenticate_refresh_enable(self) -> bool:
-        connected = self.connect()
+    def _authenticate_refresh_capabilities(self) -> None:
         if self._protocol == ProtocolVersion.V3:
             self.authenticate()
         self.refresh_status(wait_response=True)
         self.get_capabilities()
-        return connected
 
     def connect(self) -> bool:
         """Connect to device."""
@@ -234,9 +232,6 @@ class MideaDevice(threading.Thread):
                     file,
                     lineno,
                 )
-            finally:
-                if self._socket:
-                    self._socket.close()
         self.enable_device(connected)
         return connected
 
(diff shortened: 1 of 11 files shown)


midealocal/device.py
```diff
@@ -510,10 +512,9 @@ def _check_heartbeat(self, now: float) -> None:
     def run(self) -> None:
         """Run loop."""
         while self._is_run:
-            while self._socket is None:
-                if self.connect(refresh_status=True) is False:
-                    self.close_socket()
-                    time.sleep(5)
+            if not self._socket or not self.connect():
```
rokam, line 515: This looks a little bit off. It'll call connect if socket is not none and a new socket will be open.
I didn't put my thought in this but this should be something like this:
```suggestion
            if self._socket is None and not self.connect():
```

rokam (changes requested):
I saw just a possible issue.

midealocal/device.py
chemelli74: `if not self._socket` gives the same result as `if self._socket is None`

rokam: Anyway, it should use ` and` instead of `or`

chemelli74: not convinced at all, but I'll change it if you think it should be changed.

rokam: if socket is not none it'll go to connect that will replace the socket, leaving an open socket.

chemelli74: With suggested change, mypy complains:

image

chemelli74: > if socket is not none it'll go to connect that will replace the socket, leaving an open socket.

For this we can just add a : 

```python
finally:
                self._socket.close()
```

in the connect() try block

rokam: You didn't get it.

If you use `or`:
| socket is None | connect return | result |
|--------|--------|--------|
| True | Won't be run | won't raise socketexception |
| False | False | There is an active socket and another one is instantiated inside connect function, then socketexception is raised |
| False | True | There is an active socket and another one is instantiated inside connect function, no socketexception is raised |

chemelli74: Reworked to improve socket management, even if I think it's still not optimal.
Enough to proceed with this PR.

rokam (approved):
LGTM

PR #245, opened 2024-07-26
fix(cloud): fix email obfuscation

rokam:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of tokens with leading single quotes in data processing, ensuring correct functionality in regular expression substitutions.


midealocal/cloud.py
```diff
@@ -126,8 +126,13 @@ def _redact_data(data: str) -> str:
         + getattr(cr, "street_addresses", [])
     )
     for token in token_list:
-        m = len(token)
-        elm = r"\b" + token + r"\b"
+        item = token
```
chemelli74, line 129: Why not using token directly ?

rokam: to identify the begining and end of each token

chemelli74 (approved):

PR #255, opened 2024-07-31
fix: tank is always seen as full

chemelli74:
## Summary by CodeRabbit

- **New Features**
	- Enhanced the tank fullness calculation for better reliability and clarity in device management.
	- Introduced debug logging to provide insights into the tank's status.
  
- **Bug Fixes**
	- Improved the logic for updating the tank's fullness attribute to ensure accuracy based on the latest data.


Vetti52:
Tank_full is now working as expected.

chemelli74:
> Tank_full is now working as expected.

As discussed via Discord, waiting for your last test before merging.

Vetti52:
Currently I could confirm, that the status changed to „problem“, when the tank was completely filled, and/or when removed. And after inserting the empty tank again, the status changed to „ok“. So, it works as expected now.

rokam (approved):
LGTM

PR #271, opened 2024-08-12
fix: body_type default value is zero and not None

wuwentao:
sorry for the delay to fix this bug, 
it should fix https://github.com/wuwentao/midea_ac_lan/issues/176 and some other devies issue for boy_type value error when control devices.

root cause from the debug log, just use power on device as example, **all the control message should be error** after this changes:

current issue summary from the issue debug log:  
from `v0.4.1` to `v0.4.2` changes the `body_type` default value check condition casued this issue(the latest version is the same as v0.4.2)

from the debug log we can found the issue root cause is :
`v0.4.1`:  `0000000001000000800000000000ff00000000`
`v0.4.2`: `00000001000000800000000000ff00000000`

> `v0.4.2` missed `00` in  body message, after checked the source code, I found it should be `body_type` , and the default value should be `0x00`:

midea_ac_lan `v0.4.1` using midealocal `v1.0.3`
midea_ac_lan `v0.4.2` using midealocal `v1.1.2`

code changes: 
in midealocal `v1.0.3`:  https://github.com/rokam/midea-local/blob/46350f511ee0beceed8f8fb18be36a9f91378714/midealocal/message.py#L36

after changes to `v1.1.2`:
https://github.com/rokam/midea-local/blob/2f53134cd051cac3ed69162878ce3846c14cbfba/midealocal/message.py#L117

so the default value chaged from `00` to `NONE`:
in `v1.0.3`: https://github.com/rokam/midea-local/blob/46350f511ee0beceed8f8fb18be36a9f91378714/midealocal/message.py#L137

in `v1.1.2`: https://github.com/rokam/midea-local/blob/2f53134cd051cac3ed69162878ce3846c14cbfba/midealocal/message.py#L239

original code is `if self.body_type is not None:`
and after changes is `if self.body_type != NONE_VALUE:`

so there is no **body_type** default value `0x00` append to the body message and causes this issue.

as this is a global message func,  just rollback the code changes and it should append the body_type `0x00` to body message.
in addition, also rename the `NONE_VALUE` to `ZERO_VALUE` as it's not none value, it should be zero value and still need to append to body message.

please help to review and double confirm it.

Thanks

## Summary by CodeRabbit

- **New Features**
  - Updated the default value for the `body_type` parameter across multiple message classes from `NONE_VALUE` to `ZERO_VALUE`, enhancing clarity on how message body types are initialized.

- **Bug Fixes**
  - Improved consistency in message handling by standardizing the representation of default values within the Message classes, potentially reducing issues related to uninitialized states.

- **Documentation**
  - Clarified the intended use of the default value in the message initialization process, providing better understanding for developers working with message handling.

### DIFF
--- a/midealocal/devices/e6/message.py
+++ b/midealocal/devices/e6/message.py
@@ -1,7 +1,7 @@
 """Midea local E6 message."""
 
 from midealocal.message import (
-    NONE_VALUE,
+    ZERO_VALUE,
     MessageBody,
     MessageRequest,
     MessageResponse,
@@ -18,7 +18,7 @@ class MessageE6Base(MessageRequest):
             device_type=0xE6,
             protocol_version=protocol_version,
             message_type=message_type,
-            body_type=NONE_VALUE,
+            body_type=ZERO_VALUE,
         )
 
     @property
--- a/midealocal/devices/ea/message.py
+++ b/midealocal/devices/ea/message.py
@@ -3,7 +3,7 @@
 from enum import IntEnum
 
 from midealocal.message import (
-    NONE_VALUE,
+    ZERO_VALUE,
     MessageBody,
     MessageRequest,
     MessageResponse,
@@ -26,7 +26,7 @@ class MessageEABase(MessageRequest):
         self,
         protocol_version: int,
         message_type: int,
-        body_type: int = NONE_VALUE,
+        body_type: int = ZERO_VALUE,
     ) -> None:
         """Initialize EA message base."""
         super().__init__(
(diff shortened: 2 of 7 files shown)


rokam (changes requested):
Good catch. About my suggestions, we have an Enum for bodytype. So it is better to use it when it is defined instead of a default const value.  We can keep the default const for default value definitions.

rokam:
Sure.

rokam (approved):

chemelli74:
I don't get the point of this change; seems only a rename of the    var

rokam:
Before, this was the code:
`
 if self.body_type is not None: 
`

Then it was changed to;
``
 if self.body_type != 0: 
``
 
 Now it's back to 
``
  if self.body_type is not None:
``

And we missed that due to the var name

PR #276, opened 2024-08-16
chore(device): improved authentication logs

rokam:
## Summary by CodeRabbit

- **Enhancements**
	- Improved logging functionality for the authentication and device availability processes, providing better visibility and traceability of actions taken on devices.
	- The `enable_device` method has been renamed to `set_available`, with deprecation support for backward compatibility.

### DIFF
--- a/midealocal/device.py
+++ b/midealocal/device.py
@@ -245,6 +245,12 @@ class MideaDevice(threading.Thread):
             raise SocketException
         self._socket.send(request)
         response = self._socket.recv(512)
+        _LOGGER.debug(
+            "[%s] Received auth response with %d bytes: %s",
+            self._device_id,
+            len(response),
+            response.hex(),
+        )
         if len(response) < MIN_AUTH_RESPONSE:
             self.enable_device(False)
             raise AuthException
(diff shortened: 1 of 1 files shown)


midealocal/device.py
```diff
@@ -462,7 +468,11 @@ def update_all(self, status: dict[str, Any]) -> None:
 
     def enable_device(self, available: bool = True) -> None:
         """Enable device."""
-        _LOGGER.debug("[%s] Enabling device", self._device_id)
+        _LOGGER.debug(
+            "[%s] %s device",
+            self._device_id,
+            "Enabling" if available else "Disabling",
```
chemelli74, line 474: Agree, but then function named should be changed

rokam: This might cause major conflicts and lots of refactoring, as well as becoming a breaking change.

chemelli74: True but still the name is misleading and I really don't think we should leave it like it is.
Once we find a wrong thing, we should always fix it

chemelli74 (approved):

Issue #290, opened 2024-09-06
Cannot connect


rokam:
### HA core version

2024.2.2

### Current integration version

0.5.5

### Last known working integration version

0.5.3

### Device type and model

Dehumidifier 00000Q1A (161)

### Working mobile app

Home Assisstant

### The description of problem

When unpowering the dehumidifier (e.g. by a switching plug), the connection to HA is lost and will not reconnect, when power is established again.

### The logs

[homeassistant.loader] We found a custom integration midea_ac_lan which has not been tested by Home Assistant. This component might cause stability
problems, be sure to disable it if you experience issues with Home Assistant
2024-08-11 06:30:27.100 ERROR (Luftentfeuchter) [root] Uncaught thread exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.12/site-packages/midealocal/device.py", line 515, in run
    raise CannotConnect
midealocal.exceptions.CannotConnect

[https://api.github.com/repos/wuwentao/midea_ac_lan/issues/276]

PR #291, opened 2024-09-06
fix(device): prevent exception on network failure

rokam:
Fixes: #290

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method for improved socket state management, enhancing message reception handling.

- **Bug Fixes**
	- Enhanced error handling for connection issues with a retry mechanism, improving reliability during connection failures.

### DIFF
--- a/midealocal/device.py
+++ b/midealocal/device.py
@@ -527,7 +527,7 @@ class MideaDevice(threading.Thread):
 
     def _check_state(self) -> ParseMessageResult:
         if not self._socket:
-            _LOGGER.error("[%s] Socket closed", self._device_id)
+            _LOGGER.warning("[%s] Socket closed", self._device_id)
             return ParseMessageResult.ERROR
         now = time.time()
         self._check_refresh(now)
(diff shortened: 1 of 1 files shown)


midealocal/device.py
```diff
@@ -201,41 +201,35 @@ def _authenticate_refresh_capabilities(self) -> None:
     def connect(self) -> bool:
         """Connect to device."""
         connected = False
-        for _ in range(3):
```
chemelli74, line 204: I'm against an infinite loop here.

rokam: The removal of the for loop is not infinite loop.

wuwentao:
@rokam add some changes to improve it .
1. change log level from warning to debug, as all the device init connect will report a warning log after HA reboot
2. use while loop before socket is available, as `connect` only need when `self._socket is None`
3. update the `self._authenticate_refresh_capabilities()` order only when first time connect, if refresh_status and self.socket is not none, skip `self._authenticate_refresh_capabilities()`

wuwentao:
update: add try/except to all the socket send/recv, try to catch the exception and close socket, the next loop will reconnect  when socket  is none.

wuwentao:
mark to draft as the timeout and process is error, need to continue work on it. will finish it ASAP and mark to ready for review

rokam:
Replaced by #296

PR #296, opened 2024-09-19
fix(device)!: socket exception and process rebuild

### LINKED ISSUE #290: Cannot connect

wuwentao:
changes list:
1. process all the socket send/recv in try/except to prevent socket error
2. always close socket after exception matched and force reconnect to refresh
3. increase query timeout from 1 to 2, and only apply it to query cmd/msg
4. recovery socket timeout after query and recv socket msg done in each for loop.
5. rebuild process to remove the duplicate message parse.
6. rename some func and args
7. with current changes, I don't see any timeout exception in my local device testing, if there is any new error exist, we can continue improve it in future PRs.

Fixes #290 

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for device discovery, now reporting `NoSupportedProtocol` for unsupported devices.
	- Streamlined socket connection process, improving reliability and clarity in communication.
	- Introduced `MessageResult` enumeration for improved clarity in message parsing.

- **Bug Fixes**
	- Updated tests to reflect changes in exception handling, ensuring accurate responses to device compatibility issues.

### DIFF
--- a/midealocal/cli.py
+++ b/midealocal/cli.py
@@ -20,7 +20,12 @@ from midealocal.cloud import (
     get_midea_cloud,
     get_preset_account_cloud,
 )
-from midealocal.device import AuthException, MideaDevice, ProtocolVersion, RefreshFailed
+from midealocal.device import (
+    AuthException,
+    MideaDevice,
+    NoSupportedProtocol,
+    ProtocolVersion,
+)
 from midealocal.devices import device_selector
 from midealocal.discover import discover
 from midealocal.exceptions import SocketException
@@ -121,7 +126,7 @@ class MideaCLI:
                         _LOGGER.debug("Unable to connect with key: %s", key)
                     except SocketException:
                         _LOGGER.exception("Device socket closed.")
-                    except RefreshFailed:
+                    except NoSupportedProtocol:
                         _LOGGER.exception("Unable to retrieve device attributes.")
                     else:
                         _LOGGER.info("Found device:\n%s", dev.attributes)
--- a/midealocal/device.py
+++ b/midealocal/device.py
@@ -10,7 +10,7 @@ from typing import Any
 
 from typing_extensions import deprecated
 
-from .exceptions import CannotConnect, SocketException
+from .exceptions import SocketException
 from .message import (
     MessageApplianceResponse,
     MessageQueryAppliance,
@@ -28,7 +28,9 @@ from .security import (
 MIN_AUTH_RESPONSE = 20
 MIN_MSG_LENGTH = 56
 MIN_V2_FACTUAL_MSG_LENGTH = 6
-RESPONSE_TIMEOUT = 120
+SOCKET_TIMEOUT = 10  # socket connection default timeout
+QUERY_TIMEOUT = 2  # query response in 1s, 0xAC have more queries, set to 2s
+
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


rokam (changes requested):
@wuwentao this is a breaking change. It looks good. Can you please add tests to cover the changed code?

wuwentao:
> @wuwentao this is a breaking change. It looks good. Can you please add tests to cover the changed code?

thanks for your review, but I don't thinks it's a good idea to add test now, most of the process is not fullly test and verified, we still need to improve and update it, once it stable, we can start to add test for it.

wuwentao:
in addition, I don't think it can cover all, current process  should not match the finally version, it just focus on fix the socket exception  and some duplicate process rebuild, just to match the changes, especially for query socket timeout set and recovery.

I just found some new bug and push a new commit now, we may have new bug and issue with it, anyway, we can continue improve it and try to make it stable.

rokam:
> and push a new commit now, we may



> in addition, I don't think it can cover all, current process should not match the finally version, it just focus on fix the socket exception and some duplicate process rebuild, just to match the changes, especially for query socket timeout set and recovery.
> 
> I just found some new bug and push a new commit now, we may have new bug and issue with it, anyway, we can continue improve it and try to make it stable.

I think tests will help us identify those bugs. Right now, with all those changes I still don't fully understand that and if this will fix it.

rokam (approved):

wuwentao:
@rokam  thanks, but we can only add unit test, it's not real device func test.... this is the key point.
all the unit tes only can test our code logic, and it can't test any device func, all of our unit test should design to test the code logic......
so I just think we can quickly release it and check/confirm it with real device func, then we can found the result.

anyway, we just need to publish a new release with the socket error bug, we have known this bug for too many weeks and pending the fix for a long time now. and please don't worry, I can confirm I can add some unit test later, this version should not a finally or stable version, it still have more step and process need to improve.

In a company DEV release process, they should have all the device/product to run real device func test, but we don't have all the devices.....
so I just can use all of my local device to test the logic and basic func, also refer to the old logic, then publish a release to double check with our user's device, got new bug/issue and fix it, improve it.

wuwentao:
@rokam push a new commit again.
fix: when refresh_status, socket exist, and set query timeout to 2, send query to device and  recv msg, in recv_msg got socket exception, it will close socket and trigger reconnect.
so recovery socket timeout should be skip/disable when timeout/exception exist, and continue to run the reconnect job....

in this step, I think we can only do connect and no need to probe protocol in future, but no plan to change it in current MR....
as it also required more time to test and verify it.

rokam (approved):

wuwentao:
@rokam thanks, could you help to publish a new release for midea-local?

PR #306, opened 2024-09-29
chore(cloud)!: rename MSmartHome

### LINKED ISSUE #286: Rename MSSmartHome to SmartHome
App is changed since at least few months.

To avoid confusion we should rename

Note: after updating library, also Issue template on integration side should be updated

chemelli74:
Closes #286


rokam (approved):
Lgtm

PR #311, opened 2024-10-03
chore: MessageResponse debug log

chemelli74:


wuwentao (approved):

wuwentao (changes requested):
is this possible to add device id or any info to confirm which device or which msg?
if there's more devices and messages, this debug log missed this info, and we can't catch it to device or query/set message

wuwentao:
one more comments, this header/body should already exist in current device debug log:

1.  sending example:

```
2024-10-06 23:17:47.656 DEBUG (中间卧室空调) [midealocal.device] [192414535730805] Sending: {'_device_type': <DeviceType.AC: 172>, '_message_type': <MessageType.query: 3>, '_body_type': <BodyType.B5: 181>, '_message_protocol_version': 8, '_message_id': 17, '_additional_capabilities': False, 'header': 'aa0fac00000000000803', 'body': 'b501001178', 'message_type': 'query', 'body_type': 'b5', 'self': <midealocal.devices.ac.message.MessageCapabilitiesQuery object at 0x7fdb07b49e80>}, query is False

```

2. recv example:

```
2024-10-06 23:17:47.696 DEBUG (北边卧室空调) [midealocal.devices.ac] [178120884597106] Received: {'_device_type': <DeviceType.AC: 172>, '_message_type': <MessageType.query: 3>, '_body_type': <BodyType.C1: 193>, '_message_protocol_version': 8, '_header': 'aa20ac00000000000803', '_body': <midealocal.devices.ac.message.XC1MessageBody object at 0x7fdaf3199250>, '_data': 'c12101440000000000000000000000000000000001df', 'parser_list': [], 'total_energy_consumption': 0.0, 'current_energy_consumption': 0.0, 'realtime_power': 0.0, 'header': 'aa20ac00000000000803', 'body': 'c12101440000000000000000000000000000000001df', 'message_type': 'query', 'body_type': 'c1', 'self': <midealocal.devices.ac.message.MessageACResponse object at 0x7fdaf3199190>}

```

chemelli74:
> one more comments, this header/body should already exist in current device debug log:

If you want we can skip it, but the log should be near the evaluation. 
Otherwise you don't get all the needed info.
This is why I added it here, because we calculate some values and they can fail.

rokam (approved):
Lgtm

wuwentao:
I'm also try to add more debug log output, but it will caused too many debug log flood.

as current three lines output already exist in our debug log, and these header/body/message is duplicate output.

if there is some specially user really need these debug output, I think we can simple add it in a temp file and user manual replace it.

chemelli74:
@wuwentao generally speaking users should only enable debug when there is a issue. Normal level should be info.

I would also like to have a log line everywhere we do a manipulation of data so if something goes wrong, we have a easy evidence .

wuwentao:
This is not the point.
header/body/message already exist in current debug log output. we already print it.
why we need to add these duplicate info with 3 lines for every msg?

PR #313, opened 2024-10-04
fix: dedup protocol version var

chemelli74:
We have 2 variables for the same information:

- `self._protocol = protocol`
- `self._protocol_version: ProtocolVersion = ProtocolVersion.V1`

Those can diverge and checks can fail based on which one is used.


chemelli74:
Wrong assumpion based on meaningless var. Follow up #315

PR #320, opened 2024-10-05
fix: protocol check in B6 devices

chemelli74:


rokam:
Why change it without an issue related?

chemelli74:
> Why change it without an issue related?

This is the only device with a check and IMHO was based on a "device" logic not a message one

rokam:
> > Why change it without an issue related?
> 
> This is the only device with a check and IMHO was based on a "device" logic not a message one

We shouldn't change it without any related issues. We might create issues and we need stability.

chemelli74:
Anyway we need to find a proper way to handle.
I'm 99% sure that it was a mistake ; is the only device with a if on the protocol.

If we don't change the logic, we need to introduce a constant for message protocol value(s)

rokam:
Or just skip that line.

chemelli74:
Nope, we will have doubts in future

wuwentao (approved):

PR #328, opened 2024-10-16
feat(ed): add all query msg for 0xED device

wuwentao:
add all query msg for ed devices
fix https://github.com/wuwentao/midea_ac_lan/issues/352
fix https://github.com/wuwentao/midea_ac_lan/issues/402
fix https://github.com/wuwentao/midea_ac_lan/issues/424


rokam (changes requested):
Is there a reason to create another class and not to use the body_type parameter?
I prefer that you create the query message twice each one with a different body type.

wuwentao:
@rokam please check the source code from device.py with refresh_status steps.
it can't work anymore

midealocal/devices/ed/__init__.py
```diff
@@ -77,9 +83,12 @@ def _use_new_set(self) -> bool:
         # if (self.sub_type > 342 or self.sub_type == 340) else False
         return True
 
-    def build_query(self) -> list[MessageQuery]:
+    def build_query(self) -> list[MessageQuery | MessageQuery01]:
         """Midea ED device build query."""
-        return [MessageQuery(self._message_protocol_version, self._device_class)]
+        return [
+            MessageQuery(self._message_protocol_version),
+            MessageQuery01(self._message_protocol_version),
```
rokam, line 90:
```suggestion
            MessageQuery(self._message_protocol_version, ListTypes.X01),
```

midealocal/devices/ed/__init__.py
```diff
@@ -77,9 +83,12 @@ def _use_new_set(self) -> bool:
         # if (self.sub_type > 342 or self.sub_type == 340) else False
         return True
 
-    def build_query(self) -> list[MessageQuery]:
+    def build_query(self) -> list[MessageQuery | MessageQuery01]:
         """Midea ED device build query."""
-        return [MessageQuery(self._message_protocol_version, self._device_class)]
+        return [
+            MessageQuery(self._message_protocol_version),
```
rokam, line 89:
```suggestion
            MessageQuery(self._message_protocol_version, ListTypes.X00),
```

tests/devices/ed/device_ed_test.py
```diff
@@ -89,8 +89,9 @@ def test_process_message(self) -> None:
     def test_build_query(self) -> None:
         """Test build query."""
         queries = self.device.build_query()
-        assert len(queries) == 1
+        assert len(queries) == 2
         assert isinstance(queries[0], MessageQuery)
+        assert isinstance(queries[1], MessageQuery01)
```
rokam, line 94: Check body_type value here.

tests/devices/ed/message_ed_test.py
```diff
@@ -14,6 +14,7 @@
     MessageEDResponse,
     MessageNewSet,
     MessageQuery,
+    MessageQuery01,
```
rokam, line 17:
```suggestion
```

midealocal/devices/ed/message.py
```diff
@@ -77,6 +81,26 @@ def _body(self) -> bytearray:
         return bytearray([0x01])
 
 
+class MessageQuery01(MessageEDBase):
```
rokam, line 84: This class isn't needed

rokam (changes requested):
I did it with suggestions to be clear. It will work.

midealocal/devices/ed/__init__.py
```diff
@@ -77,9 +83,12 @@ def _use_new_set(self) -> bool:
         # if (self.sub_type > 342 or self.sub_type == 340) else False
         return True
 
-    def build_query(self) -> list[MessageQuery]:
+    def build_query(self) -> list[MessageQuery | MessageQuery01]:
```
rokam, line 86:
```suggestion
    def build_query(self) -> list[MessageQuery]:
```

rokam (changes requested):
Also, `self._device_class = ListTypes.X00` isn't needed anymore.

midealocal/devices/ed/__init__.py
wuwentao: if device NOT support `MessageQuery` and ONLY support `MessageQuery01`,  we only use `MessageQuery` with the same name, when device send `MessageQuery` with body 00, it will add `MessageQuery` to `self._unsupported_protocol`, so `MessageQuery` with body 01 will be ignored and can't work anymore.
this is the reason why we need to add it with a different class name

PR #329, opened 2024-10-16
feat(db): add new attributes for db device

wuwentao:
add new attributes for db device
fix https://github.com/wuwentao/midea_ac_lan/issues/335

the remaining PR will be finished in midea_ac_lan with another PR later.

### DIFF
--- a/midealocal/devices/db/__init__.py
+++ b/midealocal/devices/db/__init__.py
@@ -2,7 +2,7 @@
 
 import logging
 from enum import StrEnum
-from typing import Any
+from typing import Any, ClassVar
 
 from midealocal.const import DeviceType, ProtocolVersion
 from midealocal.device import MideaDevice
@@ -18,6 +18,18 @@ class DeviceAttributes(StrEnum):
 
     power = "power"
     start = "start"
+    status = "status"
+    mode = "mode"
+    program = "program"
+    water_level = "water_level"
+    temperature = "temperature"
+    dehydration_speed = "dehydration_speed"
+    wash_time = "wash_time"
+    wash_time_value = "wash_time_value"
+    dehydration_time = "dehydration_time"
+    dehydration_time_value = "dehydration_time_value"
+    detergent = "detergent"
+    softener = "softener"
     washing_data = "washing_data"
     progress = "progress"
     time_remaining = "time_remaining"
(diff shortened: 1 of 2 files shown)


midealocal/devices/db/__init__.py
```diff
@@ -26,6 +38,134 @@ class DeviceAttributes(StrEnum):
...
+    }
+
+    _temperature: ClassVar[dict[int, str]] = {
+        0x01: "0",
+        0x02: "20",
+        0x03: "30",
+        0x04: "40",
+        0x05: "60",
+        0x06: "95",
+        0x07: "70",
```
chemelli74, line 165: I think the 2 values are swapped, the progress is rising

midealocal/devices/db/__init__.py
```diff
@@ -26,6 +38,134 @@ class DeviceAttributes(StrEnum):
...
+        0x07: "1600",
+        0x08: "1300",
+        0xFF: "default",
+    }
+
+    _water_level: ClassVar[dict[int, str]] = {
+        0x01: "Low",
+        0x02: "Mid",
+        0x03: "High",
+        0x04: "4",
```
chemelli74, line 68: Should be "translated" to a string: `VeryHigh` ?

midealocal/devices/db/__init__.py
```diff
@@ -82,10 +234,52 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
-                    self._attributes[status] = progress[getattr(message, str(status))]
+                    self._attributes[status] = progress[value]
+                # parse status
+                elif status == DeviceAttributes.status:
+                    if value in MideaDBDevice._status:
+                        self._attributes[DeviceAttributes.status] = (
+                            MideaDBDevice._status.get(value)
+                        )
+                    else:
+                        self._attributes[DeviceAttributes.status] = None
```
chemelli74, line 248:
```suggestion
                    self._attributes[DeviceAttributes.status] = (
                         MideaDBDevice._status.get(value)
                        )
```

The purpose of `get()` is exactly to avoid the if

midealocal/devices/db/__init__.py
```diff
@@ -82,10 +234,52 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+                # parse dehydration_speed
+                elif status == DeviceAttributes.dehydration_speed:
+                    if value in MideaDBDevice._dehydration_speed:
+                        self._attributes[DeviceAttributes.dehydration_speed] = (
+                            MideaDBDevice._dehydration_speed.get(value)
+                        )
+                    else:
+                        self._attributes[DeviceAttributes.dehydration_speed] = None
```
chemelli74, line 256: Same as above

midealocal/devices/db/__init__.py
```diff
@@ -82,10 +234,52 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+                # parse water_level
+                elif status == DeviceAttributes.water_level:
+                    if value in MideaDBDevice._water_level:
+                        self._attributes[DeviceAttributes.water_level] = (
+                            MideaDBDevice._water_level.get(value)
+                        )
+                    else:
+                        self._attributes[DeviceAttributes.water_level] = None
```
chemelli74, line 264: Same as above

midealocal/devices/db/__init__.py
```diff
@@ -82,10 +234,52 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+                # parse program
+                elif status == DeviceAttributes.program:
+                    if value in MideaDBDevice._program:
+                        self._attributes[DeviceAttributes.program] = (
+                            MideaDBDevice._program.get(value)
+                        )
+                    else:
+                        self._attributes[DeviceAttributes.program] = None
```
chemelli74, line 272: Same as above

midealocal/devices/db/__init__.py
```diff
@@ -82,10 +234,52 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+                # parse temperature
+                elif status == DeviceAttributes.temperature:
+                    if value in MideaDBDevice._temperature:
+                        self._attributes[DeviceAttributes.temperature] = (
+                            MideaDBDevice._temperature.get(value)
+                        )
+                    else:
+                        self._attributes[DeviceAttributes.temperature] = None
```
chemelli74, line 280: Same as above

chemelli74 (changes requested):

midealocal/devices/db/__init__.py
wuwentao: this is the origina value from lua script.

midealocal/devices/db/__init__.py
wuwentao: this is the origina value from lua script.

midealocal/devices/db/__init__.py
wuwentao: value may not list in _status, just return None.

midealocal/devices/db/__init__.py
chemelli74: Please share the LUA script

midealocal/devices/db/__init__.py
chemelli74: ?

wuwentao: it's a very simple logic, if value not in list, we can't return error, just set the status to None.
so why you thinks the value always in _status ? we just list what we want to display, and it's not a full list with different sub modle for this device.
if there is any issue ,please add your comments, thanks

wuwentao: fixed with comment comments

wuwentao:
merge the first version and will improve it with device test result.

PR #330, opened 2024-10-16
feat(dc): add new attributes for exist dc device

wuwentao:
add new attributes for dc device
dc device should share most of the same feature as db device, add it at the same time

the remaining PR will be finished in midea_ac_lan with another PR later.


rokam (changes requested):
Looks good, can you please add unit testing?

midealocal/devices/dc/__init__.py
```diff
@@ -26,6 +36,71 @@ class DeviceAttributes(StrEnum):
...
+        3: "pause",
+        4: "end",
+        5: "prevent_wrinkle_end",
+        6: "delay_choosing",
+        7: "fault",
+        8: "delay",
+        9: "delay_pause",
+    }
+
+    _program: ClassVar[dict[int, str]] = {
```
chemelli74, line 51: Sounds similar if not identical to the  previous PR. Can we consolidate ?

midealocal/devices/dc/__init__.py
```diff
@@ -81,8 +166,26 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
-                    self._attributes[status] = progress[getattr(message, str(status))]
+                    self._attributes[status] = progress[value]
+                # parse status
+                elif status == DeviceAttributes.status:
+                    if value in MideaDCDevice._status:
+                        self._attributes[DeviceAttributes.status] = (
+                            MideaDCDevice._status.get(value)
+                        )
+                    else:
+                        self._attributes[DeviceAttributes.status] = None
```
chemelli74, line 180:
```suggestion
                    self._attributes[DeviceAttributes.status] = (
                            MideaDCDevice._status.get(value)
                        )
```

The purpose of get() is exactly to avoid the if

midealocal/devices/dc/__init__.py
```diff
@@ -81,8 +166,26 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+                # parse program
+                elif status == DeviceAttributes.program:
+                    if value in MideaDCDevice._program:
+                        self._attributes[DeviceAttributes.program] = (
+                            MideaDCDevice._program.get(value)
+                        )
+                    else:
+                        self._attributes[DeviceAttributes.program] = None
```
chemelli74, line 188: Same as above

chemelli74 (changes requested):

midealocal/devices/dc/__init__.py
wuwentao: value may not list in _status

midealocal/devices/dc/__init__.py
wuwentao: different device type have different lua protocol, we can't merge it to one.

PR #337, opened 2024-11-29
feat(e1): add new attributes for e1 device

wuwentao:
add E1 device new attributes, current only quer/read data and will continue add set/control later


wuwentao:
merge the first version to add more attrs
set/control can be done via another PRs in next release.

PR #340, opened 2024-12-13
feat(c3): add disinfect and fix tbh set error

wuwentao:
1. current disinfect is the same as tbh, fix tbh set error
2. add disinfect query
3.  TO BE DONE: set disinfect, as it's a switch
4. add some addition attrs

fix https://github.com/wuwentao/midea_ac_lan/issues/399

and https://github.com/wuwentao/midea_ac_lan/issues/404 maybe need a new PR for it.


wuwentao:
most of the code logic is ready, and test with some C3 devices, new query is not support, will merge it and check for more C3 device debug log, try to improve it in future release.

PR #353, opened 2025-02-18
feat(e6): add cold water and climate compensation features

liuwang97:
- Add cold water single/dot control attributes and methods
- Add climate compensation mode support
- Update message handling for new features


wuwentao:
@liuwang97  thank you very much for your PR.
we should fix github action error at first.
in addition, for `heating_modes`, seems there is `winter_mode` and `summer_mode` exist, maybe we can add it to current PR for different 0xE6 device.

PR #389, opened 2025-08-10
feat(a1): allow customization of default modes and speeds for dehumidifiers

agorski3:
This allows one to set custom values for the default **modes** and **speeds** of a dehumidifier. The code was largely copied from other device classes with the `set_customize` method and modified to fit this use case. I tested it without issue on my device. 

For my use case, I was trying to modify the **modes** and **speeds** in HA to match those of the Midea 50 pt. Cube Dehumidifier with Built-in Pump (MAD50PS1QWT). However, this is applicable to any other dehumidifier that has different **modes** and **speeds** than the defaults. 

For the MAD50PS1QWT device, the valid custom configuration is:

`{ "speeds": {"40": "Low", "80": "High"}, "modes": {"1": "Manual", "2": "Continuous", "4": "Max"} }`

Generally, this custom configuration should be the same for all Midea Cube Dehumidifiers and should solve the problems discussed in [Issue 570](https://github.com/wuwentao/midea_ac_lan/issues/570).

### DIFF
--- a/midealocal/devices/a1/__init__.py
+++ b/midealocal/devices/a1/__init__.py
@@ -1,5 +1,6 @@
 """Midea local A1 device."""
 
+import json
 import logging
 from enum import StrEnum
 from typing import Any, ClassVar
@@ -33,14 +34,14 @@ class DeviceAttributes(StrEnum):
 class MideaA1Device(MideaDevice):
     """Midea A1 Device."""
 
-    _modes: ClassVar[list[str]] = [
-        "Manual",
-        "Continuous",
-        "Auto",
-        "Clothes-Dry",
-        "Shoes-Dry",
-    ]
-    _speeds: ClassVar[dict[int, str]] = {
+    _default_modes: ClassVar[dict[int, str]] = {
+        1: "Manual",
+        2: "Continuous",
+        3: "Auto",
+        4: "Clothes-Dry",
+        5: "Shoes-Dry",
+    }
+    _default_speeds: ClassVar[dict[int, str]] = {
         1: "Lowest",
         40: "Low",
         60: "Medium",
@@ -61,7 +62,7 @@ class MideaA1Device(MideaDevice):
         device_protocol: ProtocolVersion,
         model: str,
         subtype: int,
-        customize: str,  # noqa: ARG002
+        customize: str,
     ) -> None:
         """Initialize Midea A1 device."""
         super().__init__(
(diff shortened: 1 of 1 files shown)


wuwentao:
@agorski3  thank you very much for your PR.
please help to fix the commit message check error.

in addition, I'm not fully understand the dict(v,k) in message.mode for set control message.

I have defined a `get_key_by_value` to do this,
you can refer to https://github.com/midea-lan/midea-local/blob/181499e35245b50d89949ea98550a710afc149b2/midealocal/devices/cd/__init__.py#L213

agorski3:
I think I correct the commit message error.

As for the the dict, I use that in several places with the same effect. So take the code below

`dict((v,k) for k,v in self._speeds.items())`

All it is doing is looping through the k,v pairs of the items in the dict _speeds and then redefining them as a tuple v,k. All these tuples are then fed to dict which creates a inverted form of the original dict where the previous keys are now the values and the previous values are now the keys. The rest of the line of code below is then just a lookup on the new dict.

`dict((v,k) for k,v in self._speeds.items())[self._attributes[DeviceAttributes.fan_speed]]`

The new dict is used to look up a key from the original dict by the original value. It is effectively the same as the method by provided just through the creation of a second inverted dict. 

This is basically the same as what is discussed here https://stackoverflow.com/questions/1031851/how-do-i-exchange-keys-with-values-in-a-dictionary

wuwentao:
@agorski3 thank you for your answer, my question should be set `message.mode` or `message.fan_speed` required a int value, and you set it to `dict(v,k)`, this should my question, sorry for my previous simple answer. 
anyway, there is a format error in per-commit ci check.

agorski3:
Ya all areas where dict is used have a lookup after them which probably got word wrapped or hidden because those two lines are long. Nothing is assigned a dict, everything is still assigned an int like before. But it should be obvious now with the changes that hopefully that fixed the style errors.

wuwentao:
> Ya all areas where dict is used have a lookup after them which probably got word wrapped or hidden because those two lines are long. Nothing is assigned a dict, everything is still assigned an int like before. But it should be obvious now with the changes that hopefully that fixed the style errors.

ok, thank you very much for your detail answer, we can check the result in pre-commit and mypy rule later via CI result.
you may have a try with local devcontainer, or just fix these error step-by-step.

wuwentao:
permission error:  100644 → 100755

wuwentao:
> Not sure how that happened or why pre-commit did not catch it on my end. Maybe because the machine I was working from yesterday was windows. Also why I could not run docker. Anyways, just spun up the dev container, ran pre-commit before and after my changes to confirm there was nothing else, ran the build successfully, and ran the tests successfully. So hopefully we are good.
> 
> Also for the future, I would recommend adding some contribution instructions to the readme. Does not have to be anything major, just general instructions on what to run to ensure we are at least passing your tests locally so you don't get bothered with commit errors. I know pre-commit takes care of some of that automatically but only if you have it installed (I did not on that machine apparently and did not notice the config file) and I had not heard of devcontainer until yesterday.

thanks for your suggestion and reply. I will try to add a contribution instructions in readme.

it works now, I will merge it and will be available in next release.
if possible, could you help to add these new feature and usage info in midea_ac_lan A1 device doc? 
github action may failed, as midealocal not publish a new release, anyway, I will fix it once midealocal publish a new release (maybe need fix some new issues/PRs, we can't pulish a new release one-by-one, so we can hold one for some days with new  issue/PRs.)

PR #411, opened 2025-11-02
fix(ac): refactor to add new binary power format

folfy:
merged AC energy / power functions
-> Rewrote these and corresponding calls and helper-functions, to make them more versatile / deduplicate common code

Fixed issue wuwentao/midea_ac_lan#678: AC energy values are x10 of real value (Midea PortaSplit)
-> Introduced format BINARY1 (power_analysis_method 12 = variant of method 2)
Lower digit still represents existing base-formats (BCD, binary, mixed/int), while higher digit(s) can be used to represent different precisions/scalars

Tested changes locally: power_analysis_method 2 and 12 (also got a Midea PortaSplit, actually needing mode 12)

Also, this fixes a bug with energy values >= 655.35 kWh (mode 12) / 6553.5 kWh (existing mode 2), line 1117:
`           return float((byte1 << 32) + (byte2 << 16) + (byte3 << 8) + byte4) / 10`
-> byte1 should be shifted 24 instead of 32 bits

Change needs to be considered in documentation in https://github.com/wuwentao/midea_ac_lan/blob/master/doc/AC.md
Also document it's meaning of interpretation, since it's helpful when migrating from Midea Smart AC (mill1000/midea-ac-py):
- 1 = BCD
- 2 = binary (with default energy/consumption multiplier, so 0.1kWh resolution)
- 3 = mixed/integer
- 12 = binary (with additional x0.1 energy/consumption factor, so 0.01kWh resolution. e.g. Midea PortaSplit)

### DIFF
--- a/midealocal/devices/ac/message.py
+++ b/midealocal/devices/ac/message.py
@@ -1068,7 +1068,7 @@ class XC1MessageBody(MessageBody):
     @classmethod
     def parse_value(cls, analysis_method: int, databytes: bytearray) -> float:
         """AC C1 message body parse value."""
-        if analysis_method not in PowerFormats:
+        if analysis_method not in PowerFormats._value2member_map_:
             return 0.0  # unknown method
         analysis_function = cls.power_analysis_methods[analysis_method % 10]
         value = 0
--- a/tests/devices/ac/message_ac_test.py
+++ b/tests/devices/ac/message_ac_test.py
@@ -652,12 +652,15 @@ class TestMessageACResponse:
         assert hasattr(response, "realtime_power")
         assert response.realtime_power == expected_realtime_power
 
-    def test_message_query_c1_method2(self) -> None:
-        """Test Message parse query C1 method2."""
+    def test_message_query_c1_method2(self, method: int = 2) -> None:
+        """Test Message parse query C1 method2 (and 12)."""
         self.header[9] = 0x03
         body = bytearray(20)
         body[0] = 0xC1  # Body type
         body[3] = 0x44  # Set the type to 0x44
+
+        # method 12 is like 2, but with 0.01kWh resolution instead of 0.1kWh
+        energy_divisor = 10 if method == 2 else 100
 
         # Total energy consumption bytes
         body[4] = 0x01
@@ -665,7 +668,7 @@ class TestMessageACResponse:
         body[6] = 0x45
         body[7] = 0x67
         expected_total_energy = (
-            float((0x01 << 32) + (0x23 << 16) + (0x45 << 8) + 0x67) / 10
+            float((0x01 << 24) + (0x23 << 16) + (0x45 << 8) + 0x67) / energy_divisor
         )
 
         # Current energy consumption bytes
@@ -674,7 +677,7 @@ class TestMessageACResponse:
         body[14] = 0xCD
         body[15] = 0xEF
         expected_current_energy = (
-            float((0x89 << 32) + (0xAB << 16) + (0xCD << 8) + 0xEF) / 10
+            float((0x89 << 24) + (0xAB << 16) + (0xCD << 8) + 0xEF) / energy_divisor
         )
 
         # Real-time power bytes
@@ -683,7 +686,7 @@ class TestMessageACResponse:
         body[18] = 0x56
         expected_realtime_power = float((0x12 << 16) + (0x34 << 8) + 0x56) / 10
 
-        response = MessageACResponse(self.header + body, 2)
+        response = MessageACResponse(self.header + body, method)
 
         assert hasattr(response, "total_energy_consumption")
         assert response.total_energy_consumption == expected_total_energy
@@ -691,6 +694,10 @@ class TestMessageACResponse:
         assert response.current_energy_consumption == expected_current_energy
         assert hasattr(response, "realtime_power")
         assert response.realtime_power == expected_realtime_power
+
+    def test_message_query_c1_method12(self) -> None:
+        """Test Message parse query C1 method12."""
+        self.test_message_query_c1_method2(12)
 
     def test_message_query_c1_method3(self) -> None:
         """Test Message parse query C1 method3."""


folfy:
Well, TL;DR;, I fixed the typing and formatting issues, and also some old bugs, like a bad length check (same variable was assigned twice), and removed some redundant/unnecessary code (as highlighted by ruff / mypy / VSCode / etc and confirmed to be OK).

@wuwentao I've spent like at least an hour trying to setup VSCode on my Win 11, with all the plugins and everything, but it was a pain to get that all running with the devcontainer running here.

Checking your Dockerfile, it seems to be related to your pip mirror configuration, because of which it attempts connecting to **-pypi.org** (leading dash):
 ```
> [dev_container_auto_added_stage_label 6/7] RUN pip3 --disable-pip-version-check --no-cache-dir install -r /tmp/pip-tmp/requirements-all.txt    && rm -rf /tmp/pip-tmp:
0.497 Looking in indexes: https://-pypi.org/simple                              
0.501 WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x72bb11003830>: Failed to establish a new connection: [Errno -2] Name or service not known')': /simple/aiofiles/
```

**Dockerfile** - looks OK:
```
RUN echo "[global]" > /etc/pip.conf && \
    echo "index-url = https://${PIP_MIRROR_DOMAIN}/simple" >> /etc/pip.conf && \
    echo "trusted-host = ${PIP_MIRROR_DOMAIN}" >> /etc/pip.conf
```


**devcontainer.json** - Idk how parameter expansion works when VSCode loads devcontainer.json, like I know bash would substitute `${<varname>:-<default-value>}`, but here the dash in front of pypi.org clearly seems preserved:
`"PIP_MIRROR_DOMAIN": "${localEnv:PIP_MIRROR_DOMAIN:-pypi.org}"`
I'm also not sure what the first `localEnv:` is there for / intended to do, like in bash this wouldn't rly make sense, but this should be fixed somehow. Correcting this locally with a static configuration gets me just one step further:

```
Running the postCreateCommand from devcontainer.json...

[31195 ms] Start: Run in container: /bin/sh -c scripts/setup.sh
/usr/bin/env: ‘bash\r’: No such file or directory
```

So a Windows CRLF is causing trouble - Found the shebang in script/setup.sh and changed that file from CRLF to LF.

```
Running the postCreateCommand from devcontainer.json...

[8987 ms] Start: Run in container: /bin/sh -c scripts/setup.sh
pre-commit installed at .git/hooks/pre-commit
pre-commit installed at .git/hooks/commit-msg

added 11 packages in 2s
npm notice
npm notice New patch version of npm available! 11.6.1 -> 11.6.2
npm notice Changelog: https://github.com/npm/cli/releases/tag/v11.6.2
npm notice To update run: npm install -g npm@11.6.2
npm notice
Done. Press any key to close the terminal.
```

Funny enough, all my files are marked as modified now suddenly, and ofc it's the Windows CRLF line-endings everywhere. Don't know what changed the core.autocrlf setting, would have needed that before already, but it's something to be mentioned for other devs (making sure they check this out with LF). Maybe you could consider writing up a brief CONTRIBUTING.md, with what all you expect to be setup/run and how, before creating a PR?

**Comment by wuwentao** [comment]
@folfy please check with this PR about docs:
https://github.com/midea-lan/midea-local/pull/413
if there is any error exist, I will update it.
once got confirmed, i will merge it.
thanks

**Comment by wuwentao** [comment]
There is some testcases for 0xAC device.
please help to check the error and try to fix it.
thanks!

**Comment by folfy** [comment]
## VSCode/devcontainer setup and proxy config

> @folfy sorry, as most of us under GFW in china network, we can't directly access github/pip/docker/etc, so we need to config a local mirror about these sites: ` export APT_MIRROR_DOMAIN=xxxxx`, or redirect these domain to proxy.
> 
> for windows user, **WSL2** should be the expected solution with VScode + devcontainer. install a ubuntu under windows WS2, all these feature should works. finally, I will try to add some info in CONTRIBUTING.md.

Yeah, I already had WSL2 with Ubuntu installed, but still ran into all these issues mentioned above. At least the devcontainer.json should be fixed:
`"PIP_MIRROR_DOMAIN": "${localEnv:PIP_MIRROR_DOMAIN:-pypi.org}"`
I found out that "localEnv:PIP_MIRROR_DOMAIN" is a docker (compose?) directive to get this from the users environment. Not sure where / how the default value is expanded, but removing the dash after the colon fixes the issue (using the last value as default), so proposing that fix for devcontainer.json:7-8:
```
      "APT_MIRROR_DOMAIN": "${localEnv:APT_MIRROR_DOMAIN:deb.debian.org}",
      "PIP_MIRROR_DOMAIN": "${localEnv:PIP_MIRROR_DOMAIN:pypi.org}"
```

## Documentation

> @folfy please check with this PR about docs:
> #413
> if there is any error exist, I will update it.
> once got confirmed, i will merge it.
> thanks

Thanks, yeah, I'll check later on today or tomorrow 👍 

## Tests

> There is some testcases for 0xAC device. please help to check the error and try to fix it. thanks!

### Pylint config issue

Unfortunately tests don't seem to run here - I think cause I got some other Pylint error here still, that I didn't bother to investigate before:
```
Unrecognized option found: suggestion-mode Pylint(E0015:unrecognized-option)
```
Seems your setting pylintrc:25 `suggestion-mode=yes` has been deprecated in Pylint 4.0:
https://pylint.readthedocs.io/en/latest/whatsnew/4/4.0/index.html
So I removed that locally here now.

### ruff issue

Also this (typical Windows) issue raised by ruff is probably still blocking my tests:
```
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1

midealocal/devices/ac/message.py:1:1: EXE002 The file is executable but no shebang is present
Found 1 error.
```
Guess it's copying the checkout from Windows/NTFS to WSL, where NTFS doesn't have the executable flag, and correspondingly flags every file executable when copying or mount. Dunno how to fix that yet. Ignoring it though VSCode doesn't seem to work... @wuwentao  Are you running VSCode natively on Windows? I usually develop on my headless server with vim  and co., so not bothering with a graphical environment there. Thought if I need to use VSCode I best run it on Windows, but seems to cause a lot of issues.

### Failed test

Yes, this test also is flawed, just like the function itself was:

> Also, this fixes a bug with energy values >= 655.35 kWh (mode 12) / 6553.5 kWh (existing mode 2), line 1117:
> ` return float((byte1 << 32) + (byte2 << 16) + (byte3 << 8) + byte4) / 10`
> -> byte1 should be shifted 24 instead of 32 bits

Since I fixed that, I'll had to correct the two tests in message_ac_test.py line 668 and 677 (shift the highest byte only by 3 bytes / 24 bits instead of 32). Also created a new test-case for method 12, re-using the test for method 2. Hope it's good now, due to the ruff issue above I still can't run tests... 😞

PR #420, opened 2025-12-28
fix(CD): updating to inconsistent values and missing F to C conversion

FrancYescO:
https://github.com/wuwentao/midea_ac_lan/issues/742 https://github.com/wuwentao/midea_ac_lan/issues/734

### DIFF
--- a/midealocal/devices/cd/__init__.py
+++ b/midealocal/devices/cd/__init__.py
@@ -106,13 +106,13 @@ class MideaCDDevice(MideaDevice):
         self._fahrenheit: bool = False
         self.set_customize(customize)
 
-    def _value_to_temperature(self, value: float) -> float:
+    def _value_to_temperature(self, value: float, force_fahrenheit: bool, force_old: bool) -> float:
         # fahrenheit to celsius
-        if self._fahrenheit:
-            return self.fahrenheit_to_celsius(value)
+        if self._fahrenheit or force_fahrenheit:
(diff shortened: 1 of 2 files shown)


FrancYescO:
ok another try... 

btw i'm near to fix also the set operations for CD, idk if you want to wait i'll add in this same PR

wuwentao:
> ok another try...
> 
> btw i'm near to fix also the set operations for CD, idk if you want to wait i'll add in this same PR

no problem, we have enough time to test and check with it.

FrancYescO:
@wuwentao more or less now also the set operations are working, sometime it will still fall in the 0°C bug but is still working a lot better than before... if possible can i ask you to add the lua file (like the other one `lua/T_0000_CD_14.lua` that idk from where it was retrived) for my device (SN is in the issue)?

wuwentao:
@FrancYescO 
please check with this file source code and cli tool help message, you can got the result.
if you need any help, we can sync with the future issue.
thanks

FrancYescO:
Ok thanks, if you are able to retrieve, as said can be useful to have the lua files.

wuwentao:
[0xCD_lua.zip](https://github.com/user-attachments/files/26012427/0xCD_lua.zip)

@FrancYescO 
please check with this two lua file, as my account can't login to EU or smart home, so I just download a chinese local lua file, but most of the feature/attr should be the same one.
if there is some error with your device ,we can check and confirm it later.

PR #432, opened 2026-05-21
fix: mypy check bytes and bytearray error

wuwentao:
fix mypy error with bytes and bytearray after mypy version upgrade


chemelli74 (approved):
Finally all green !

PR #448, opened 2026-06-04
feat(cc): support 0xFE VRF panel protocol (171PNL01/171PANEL)

DeemoTong:
## Problem

Newer Midea VRF indoor-unit Wi-Fi panel controllers — models **171PNL01** and **171PANEL** (device type `0xCC`) — do not work with the current CC implementation:

- `indoor_temperature` is reported as **-20 °C**, `mode` is stuck on **auto**, and `power` is always reported **on**
- the device **cannot be controlled** — changing mode or temperature makes the unit beep but nothing actually changes

These controllers are common on Midea multi-split / VRF systems, in particular the **MIH-series indoor units widely installed in Australia** (e.g. `MIH28GHN18`), where each indoor unit has its own Wi-Fi module and is normally driven by an IR remote. The same symptoms are reported in wuwentao/midea_ac_lan#691 and wuwentao/midea_ac_lan#780.

## Root cause

These devices answer the `X01` query with a **different payload layout**, flagged by a `0xFE` format byte at `body[1]`. The existing decoder treats `body[1]` as power/mode (so power/mode are always wrong) and reads the indoor temperature from `body[4]`, which is `0` on these devices → `-20 °C`. They also use a **key-value (TLV) control protocol** instead of the legacy C3 command.

## Changes

Everything is gated on the `0xFE` format byte, so **legacy CC devices are completely unaffected**:

- Decode the `0xFE` payload: power (`body[8]`), target temperature (`body[11]`), indoor temperature (`body[12:13]`, 0.1° units), mode (`body[31]`), fan speed (`body[34]`), eco (`body[56]`), sleep (`body[60]`) and swing (`body[41]`).
- Add `MessageFEControl`, a key-value control frame (2-byte id + length + value + `0xFF`, followed by a message id and a CRC8), reusing the existing `crc8.calculate` helper.
- Route control and the resulting state read-back to the new path only for `0xFE` devices.
- Add unit tests for the new decoding and the control frame.

## Attribution

The `0xFE` byte positions and the TLV control framing were reverse-engineered by the **msmart-ng** project ([mill1000/midea-ac-py](https://github.com/mill1000/midea-ac-py), MIT-licensed). No code was copied — this is an independent implementation using this library's own `crc8` helper.

## Testing

- New unit tests in `tests/devices/cc/` pass.
- Full suite: **177 passed**; `ruff` and `mypy` clean.
- Verified live on a 1-outdoor / 4-indoor VRF system (4 × 171PNL01 / 171PANEL): temperature and state read-back, plus power / mode / target-temperature / fan-speed / eco / sleep / swing control, all confirmed correct.

### DIFF
--- a/midealocal/devices/cc/__init__.py
+++ b/midealocal/devices/cc/__init__.py
@@ -134,6 +134,7 @@ class MideaCCDevice(MideaDevice):
         _LOGGER.debug("[%s] Received: %s", self.device_id, message)
         if getattr(message, "is_fe_format", False):
             self._is_fe_format = True
+            _LOGGER.debug("[%s] _is_fe_format option True", self.device_id)
         new_status = {}
         fan_speed: int | None = None
         for status in self._attributes:
--- a/midealocal/devices/cc/__init__.py
+++ b/midealocal/devices/cc/__init__.py
@@ -7,7 +7,14 @@ from typing import Any, ClassVar
 from midealocal.const import DeviceType, ProtocolVersion
 from midealocal.device import MideaDevice
 
-from .message import MessageCCResponse, MessageQuery, MessageSet
+from .message import (
+    INDEX_TO_FE_MODE,
+    CCControlId,
+    MessageCCResponse,
+    MessageFEControl,
+    MessageQuery,
+    MessageSet,
+)
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


wuwentao:
Thank you very much for your PR, it really looks good, and there is only a debug purpose output review comments, I have add it.

we can merge it and try to test and confirm with more users.
once it done, I will publish a new release with midea-local, then sync it to midea_ac_lan later.

wuwentao:
I have manual edit current PR and add one line for debug purpose, merged, thanks

PR #466, opened 2026-06-08
feat(fa): add humidify, waterions, display_on_off (AAF10MR, subtype 0)

arbaleast:
## Summary

Add protocol-level support for three FA (tower fan) features: humidify, waterions, display_on_off.

## Target device

- **Model**: AAF10MR (Tower Fan)
- **Device type**: 0xFA, subtype 0 (X00)
- **Lua codec**: T_0000_FA_5600119Z_2022072801.lua
- **Protocol**: V2/V3

Byte positions verified against Midea Cloud Lua codec and tested with real hardware.

### `message.py` (+21 lines)
- `MessageSet` fields: humidify, waterions, display_on_off (bool | None)
- SET body byte encoding for all three attributes
- Response parsing in `FAGeneralMessageBody`

### `__init__.py` (+6 lines)
- `DeviceAttributes` enum entries
- Initial attribute dict entries (default: False)

## Byte positions (subtype 0)

| Feature | SET byte | RESP byte | Lua getBits | ON val | OFF val |
|---------|----------|-----------|-------------|--------|---------|
| humidify | body[8] | body[9] bits 4-7 | msg[19] | 0x20 | 0x10 |
| waterions | body[33] | body[34] bits 0-1 | msg[34] | 0x01 | 0x02 |
| display | body[18] | body[19] bits 6-7 | msg[29] | 0x40 | 0x80 |

Lua msg[N] indexes the full Midea protocol message (10-byte header + body). Python body[N] includes the leading body_type byte at index 0.

## Compatibility

Likely compatible with all FA (tower fan) models with subtype 0 that support these features.

Confirmed on **AAF10MR**:
- humidify: works via LAN
- waterions: works via LAN
- display_on_off: state parses correctly from LAN response; LAN SET may be ignored on some models (firmware only processes display SET through MQTT/cloud path)

## HA usage

The midea_ac_lan component (midea_devices.py) already has entity mappings for these attributes. After upgrading midea-local, enable the switches via:
Settings -> Devices & Services -> Midea AC LAN -> Configure -> Switches


wuwentao:
## Code review — `feat(fa): add humidify, waterions, display_on_off`

Reviewed the encode/decode logic beyond the lint fixes. The bit-level protocol work is **correct** — I verified every new attribute's byte offset and on/off encoding against the existing attributes, and the SET↔GET mapping is consistent (see table below). I did find one real **correctness bug** (buffer overrun → `IndexError`), which I've pushed a fix for, plus a couple of observations.

### ✅ What's correct
The SET/GET byte offsets follow the established `+1` convention, and the on/off bit values round-trip cleanly:

| attribute | SET byte | GET byte | on/off encode | GET check | symmetric |
|-----------|----------|----------|---------------|-----------|-----------|
| humidify | `[8]` hi-nibble | `[9]` | on=`0x2`, off=`0x1` | `== 2` | ✅ |
| display_on_off | `[18]` bits7-6 | `[19]` | on=`0x40`(1), off=`0x80`(2) | `== 1` | ✅ |
| waterions | `[33]` low-2 | `[34]` | on=`0x01`, off=`0x02` | `== 1` | ✅ |

### 🐛 Bug fixed (pushed as `ebdeb58`): unguarded byte access → `IndexError`

The new reads/writes use fixed offsets without a length check, unlike the neighbouring `tilting_angle` code which guards every high-offset access. Two crash paths:

1. **SET side** — `MessageSet._body` builds an **18-byte** buffer for subtypes `1..10` and `0xA1`, and a **49-byte** buffer otherwise. Writing `display_on_off` (byte 18) or `waterions` (byte 33) on the 18-byte buffer raises `IndexError`. So toggling those attributes on any FA device whose subtype isn't 0 crashes. (This PR targets subtype 0, which uses the 49-byte buffer, so it works for the tested device — but the code path is reachable for others.)

2. **GET side** — `FAGeneralMessageBody` decoded `body[9]`/`body[19]`/`body[34]` unconditionally. Response bodies can be shorter than 35 bytes — the existing `tilting_angle` read is already guarded with `len(body) > 25`, which proves short bodies occur. Reading `body[34]` on a short body raises `IndexError` and breaks status decoding for **existing** FA devices, not just new ones.

**Fix** mirrors the file's own `tilting_angle` guard convention — length checks on the writes, and conditional reads defaulting to `False` (the device default). I verified with a local repro: the old code raises `IndexError` on `[33]` with the 18-byte buffer; the guarded code no longer does, and a full 49-byte body still decodes all three attributes correctly. Bit layout is unchanged — only bounds guards added.

### 💡 Observations (not blocking)
- **No test coverage.** There's no `tests/devices/fa/` at all, so the encode/decode (and this offset bug) is untested. A small round-trip test — build a `MessageSet`, read back through `FAGeneralMessageBody`, assert the three attributes — would guard the byte offsets against future regressions. Happy to add one if you'd like.
- **Unsupported-device reporting:** the device inits `humidify/waterions/display_on_off = False`, but the GET path now decodes a value for *every* FA response. On a non-humidifier FA model, bytes 9/19/34 may carry unrelated data and surface spurious states. If these features are model-specific, consider gating the decode by subtype/capability. (Low priority — worth a look if you get odd reports from non-humidifier fans.)

---
Overall: solid protocol work, one buffer-safety bug fixed. CI (ruff) is green.

PR #444, opened 2026-05-29
feat(ac): add anion, sound, self_clean, pmv and error_code support

tilalx:
# Description
Add support for anion (ionizer), sound (beeping on setting change), self_clean, PMV sensor and error code attributes for Midea AC devices.

  ---
  ### Protocol details
  
  Anion (ionizer)
  - State is read from the C0/A0 response body byte 9, bit 0x20 (the purifierValue flag already parsed but not exposed)
  - Controlled via MessageGeneralSet body byte 8, bit 0x20 — same byte as dry/aux_heating/eco_mode
  - Setting via new-protocol B0 tag 0x021E is silently ignored by the firmware; the general set path is required

  Sound
  - Uses pre-existing tag 0x022C (buzzer_all) already defined in NewProtocolTags but never queried, parsed or exposed.  Wired up query, response parsing and set support

  Self-clean
  - Triggers the self-cleaning cycle (fan dries evaporator, then unit shuts down)
  - Set-only via new-protocol tag 0x0039 (self_clean), value 0x01 to trigger / 0x00 to cancel
  - No "in progress" status bit is returned by the device; attribute stays False after polling

  PMV (Predicted Mean Vote)
  - Thermal comfort index per ASHRAE standard, range −3.5 (cold) to +3.5 (hot), 0 = neutral
  - Already parsed in XC0MessageBody and XA0MessageBody but never exposed
  - Read from C0 body byte 14 low nibble: (byte & 0x0F) * 0.5 - 3.5

  Error code
  - Device fault code; 0 = no fault, non-zero = firmware error code
  - Queried via new-protocol tag 0x003F (error_code_query) added to MessageNewProtocolQuery
  - Parsed from B1 response in XBXMessageBody

  ###  Changes

  - Add anion, sound, self_clean, pmv, error_code to DeviceAttributes
  - Add anion field to MessageGeneralSet with bit 0x20 in body byte 8
  - Add anion attribute to XC0MessageBody and XA0MessageBody (alongside existing purifier)
  - Add sound and error_code parsing to XBXMessageBody
  - Add sound, self_clean fields to MessageNewProtocolSet with pack support
  - Add buzzer_all and error_code_query to MessageNewProtocolQuery tag list
  - Route sound and self_clean through make_newprotocol_message_set in set_attribute
  - Route anion through make_message_set / general set path in set_attribute

 ### Context
  
  Capability presence confirmed via B5 capability query on a live device:
  - b5_anion (0x021E) = 1 → ionizer supported
[...]

PR #467, opened 2026-06-09
feat(ac): reflect self-clean running state from device report

tilalx:
Follow-up to #444 — self-clean running state
  
  The previous PR noted that self_clean had no "in progress" status bit. After further investigation with a live device, tag 0x00E2 (b5_self_clean_active) carries exactly that: the actual running state
  of the self-clean cycle, pushed periodically as a B5 notify2 message.

  This PR wires it up so self_clean reflects reality rather than just the last command sent.

  Protocol details

  - Tag 0x00E2 is returned in B5 push messages (message type notify2) and also queryable via NewProtocolQuery
  - The value is a 39-byte payload; byte 12 is non-zero while the cycle is running, 0x00 when idle
  - The same tag name b5_self_clean_active already exists logically in the protocol — it just wasn't queried or parsed

  Behaviour after this change

  - Setting self_clean = True sends the trigger command (unchanged from #444)
  - The device then pushes b5_self_clean_active reports; self_clean updates in real time to match
  - When the cycle ends naturally or is cancelled, the device reports 0x00 and self_clean resets to False automatically
  - No separate self_clean_active attribute is introduced — self_clean serves as both control and live status

  Changes

  - Add b5_self_clean_active = 0x00E2 to NewProtocolTags
  - Include it in MessageNewProtocolQuery
  - Parse byte 12 of the B5 response in XBXMessageBody into self_clean_active on the message object
  - In process_message, map message.self_clean_active → DeviceAttributes.self_clean

  Testing

  Verified on a PortaSplit:
  - self_clean goes True within ~6 s of triggering (first notify2 push)
  - self_clean goes False when the cycle completes or is cancelled
  - Behaviour confirmed both when starting from device-OFF and device-ON states

### DIFF
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -292,6 +292,10 @@ class MideaACDevice(MideaDevice):
             self._fresh_air_version = DeviceAttributes.fresh_air_1
         elif self._attributes[DeviceAttributes.fresh_air_2] is not None:
             self._fresh_air_version = DeviceAttributes.fresh_air_2
+        if hasattr(message, "self_clean_active"):
+            active = message.self_clean_active
+            self._attributes[DeviceAttributes.self_clean] = active
+            new_status[DeviceAttributes.self_clean.value] = active
         return new_status
 
     def make_message_set(self) -> MessageGeneralSet:
--- a/midealocal/devices/ac/message.py
+++ b/midealocal/devices/ac/message.py
@@ -31,6 +31,7 @@ FROST_PROTECT_MIN_LENGTH = 22
 INDIRECT_WIND_VALUE = 0x02
 MAX_MSG_SERIAL_NUM = 254
 OUT_SILENT_VALUE = 0x03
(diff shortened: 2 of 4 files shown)


wuwentao:
thanks for your PR and help, I have no review comments, it looks good to me.
I will merge it and try to publish it in a new release and sync it to midea-local.
in addition, could you help to submit a new PR in midea_ac_lan side, to enable this new attr and add it to docs?
and this PR will pass CI jobs after new  midea-local lib sync to midea_ac_lan,there is some CI/ENV changes and I will have some delay to finish it.
Thanks.

tilalx:
Thanks @wuwentao. I don't think a separate attribute or new PR is needed on the midea_ac_lan side here — this is intentionally linked back to self_clean itself. With this change self_clean now carries the realtime running state from the device (via the b5_self_clean_active push) instead of only reflecting the trigger command. So no new attribute to wire up: self_clean just becomes accurate on its own once midea-local is synced.

PR #469, opened 2026-06-10
feat(ac): add rate_select (power rate limit / Gen mode) support

vaughnitem:
## Summary

Exposes the `rate_select` attribute (protocol tag `0x0048`) for AC devices, enabling Home Assistant (and other consumers) to read and set the power rate limiter — commonly known as **Gen mode** or **inverter/generator mode** — which caps the unit's wattage draw.

### What is rate_select?

`rate_select` limits the AC's power consumption to a fraction of its rated capacity. This is used when running the unit on a generator or limited power source. The tag `0x0048` already existed in `NewProtocolQuery` (query-only enum) but was absent from `NewProtocolTags` (query+response), so it was never queried, parsed, or exposed.

Device capability `b5_electricity` (tag `0x0216`) advertises support; a value of `4` means 4 rate levels are available.

### Changes

**`midealocal/devices/ac/message.py`**
- Add `rate_select = 0x0048` to `NewProtocolTags`
- Add `NewProtocolTags.rate_select` to `MessageNewProtocolQuery._body` query list
- Add `self.rate_select: int | None = None` field to `MessageNewProtocolSet`
- Add rate_select TLV packing in `MessageNewProtocolSet._body`
- Parse `rate_select` from B1 responses in `XBXMessageBody`

**`midealocal/devices/ac/__init__.py`**
- Add `rate_select = "rate_select"` to `DeviceAttributes`
- Add `_rate_selects: ClassVar[dict[int, str]]` class variable mapping raw integer values to string labels (`1`, `20`, `40`, `60`, `80`, `100`)
- Add `rate_selects` property returning the options list
- Add `DeviceAttributes.rate_select: None` to initial attributes dict
- Handle `rate_select` lookup in `process_message` (int → string label via `_rate_selects`)
- Route `rate_select` through `make_newprotocol_message_set` in `set_attribute`
- Add `rate_select` branch in `make_newprotocol_message_set` (string label → int via `get_dict_key_by_value`)

### Observed values (tested on real device)

| Value | Meaning |
|-------|---------|
| `1`   | Gen mode — lowest limit (~400 W) |
| `20`  | Low power limit |
| `40`–`80` | Intermediate levels |
| `100` | No limit (full power) |

Values mirror the existing fan speed scale. The mapping was confirmed on a Midea V3 AC with `b5_electricity = 4`.

### Testing

Verified on a live Midea AC (protocol V3, token/key auth) via `midea_ac_lan` Home Assistant integration:
- B1 query response correctly includes tag `0x0048`
- State updates reflect remote changes in real time
- Set command successfully changes the rate level on the device


wuwentao:
I'm sorry, there is some github action error, and I will delay this PR merge.
as there is more ci jobs changes, if there is some github action related issue, I will try to fix it later.

wuwentao:
## Review + CI fix — `feat(ac): add rate_select (power rate limit / Gen mode)`

### Root cause of the pytest failure
`test_new_protocol_query_body` was a **stale test assertion**, not a code bug. This PR correctly adds `NewProtocolTags.rate_select` (`0x0048`) to `MessageNewProtocolQuery._body`'s query list (so the device is asked to report its rate_select state), which bumps the query-param count byte `0x0C → 0x0D` and inserts the tag pair. The test's `expected_body` was still the pre-feature list, so it mismatched:

```
At index 1 diff: 13 != 12   # count byte: actual 0x0D, expected 0x0C
Left contains 2 more items  # the rate_select tag (0x0048) the test didn't expect
```

**Fix (pushed as `630f0e9`):** updated `expected_body` to match `_body` — count byte `0x0C → 0x0D`, and inserted the `rate_select` tag pair between `wind_ud_angle` and `out_silent` (mirroring the order in `_body`). Verified locally:
- The previously-failing test now **passes**
- Full suite: **242 passed** (was 241 passed / 1 failed)
- `ruff check` on the changed files: **clean** (incl. the `# noqa: C901` from the earlier fix)

### Correctness review — the rate_select round-trip is sound ✅
I traced the full encode/decode path; it's symmetric and mirrors the existing `wind_lr_angle`/`wind_ud_angle` handling it sits beside:

| step | code | direction |
|------|------|-----------|
| decode (raw→str) | `process_message`: `_rate_selects.get(value)` | int → "20" |
| decode read | `params[rate_select][0]` → `self.rate_select` | raw int |
| encode (str→raw) | `get_dict_key_by_value("_rate_selects", str(value))` | "20" → int |
| encode pack | `_body`: `bytearray([int(self.rate_select)])` | raw int |
| query | `rate_select` added to `MessageNewProtocolQuery` params | device reports it |

- `_rate_selects` dict is consistent (`{1:"1", 20:"20", ...}`), and `set_attribute` correctly routes `rate_select` through the new-protocol set list.
- Bit/byte handling matches convention — no offset or masking issues.

### Observations (non-blocking)
- **Device-level test coverage:** there's a message-body test (now fixed), but no `device_ac_test.py` test that sets/reads `rate_select` end-to-end. This is consistent with how the neighboring angle attributes are covered, so not a blocker — but a small set/get round-trip test would be a nice follow-up.

### Verdict
Production logic is correct; the only issue was the stale test assertion, now fixed and verified green. **Ready to merge** once CI reruns confirm (I ran the full suite locally: 242/242 passing, ruff clean).

rokam:
@vaughnitem we need to add some kind of guard for rate_select. So we just query it on supported devices. It is causing issues with other devices.
In the PR, you've mentioned b5_eletricity to check the capabilities. I don't have this type of device, can you make some kind of guard for this?

This is the issue I'm refering to: https://github.com/wuwentao/midea_ac_lan/issues/963

vaughnitem:
> capabilities

I see the issue. a logic gate could definitely fix that however I have no ability to test the implementation against actual device (supported/unsupported) as I'm currently on vacation.

chemelli74:
I think we should add a guard to fix the issue so most users have a working solution and once @vaughnitem is back he can validate it works for his device as well.

Whay you think ?

rokam:
Sure, but the main issue is how to correctly identify the support. We do query `b5_eletricity` before trying to query `rate_select`? Will this cause issue with the other devices?

chemelli74:
> Sure, but the main issue is how to correctly identify the support. We do query `b5_eletricity` before trying to query `rate_select`? Will this cause issue with the other devices?

Yup I think so.
Available to test on my device from tomorrow evening.

vaughnitem:
https://github.com/midea-lan/midea-local/pull/632 new pull request has been submitted that should rectify this. all tests. passed but wasn't able to test against actual device but should work.

PR #475, opened 2026-06-20
feat(e1): add dishwasher work controls

riccihenrique:
## Summary
- Expose E1 work modes through a public modes property.
- Add set_work_mode and start_work controls backed by MessageWork.
- Add E1 device and message coverage.

## Testing
- pytest -q (247 passed)
- ruff check midealocal/devices/e1 tests/devices/e1
- ruff format --check midealocal/devices/e1 tests/devices/e1
- mypy midealocal/devices/e1 tests/devices/e1

This provides the device-message/control layer required by the related Home Assistant integration PR.


midealocal/devices/e1/__init__.py
```diff
@@ -137,10 +138,66 @@ def __init__(
...
+
+        Raises
+        ------
+        ValueError
+            If the mode code is not supported.
+
+        """
+        if mode not in self._modes:
+            msg = f"[e1] Unsupported work mode: {mode}"
+            raise ValueError(msg)
```
wuwentao: as this python lib only works in HA Core, so we may need to use `_LOGGER` with different error/warning level and message.

wuwentao:
@riccihenrique thank you very much for your PR.
Could you help to update the `raise` error with `_LOGGER` in the py file?

midealocal/devices/e1/__init__.py
```diff
@@ -137,10 +138,66 @@ def __init__(
...
+    @property
+    def modes(self) -> dict[int, str]:
+        """Return supported work modes.
+
+        Returns
+        -------
+        Mapping of mode codes to names.
+
+        """
+        return self._modes.copy()
```
chemelli74, line 135: Why copy() ?

midealocal/devices/e1/__init__.py
```diff
@@ -137,10 +138,66 @@ def __init__(
...
+        """
+        if mode not in self._modes:
+            msg = f"[e1] Unsupported work mode: {mode}"
+            raise ValueError(msg)
+
+        message = MessageWork(self._message_protocol_version)
+        message.mode = mode
+        self.build_send(message)
+
+    def start_work(self) -> None:
```
chemelli74, line 158: Why not add a parameter with the work mode ?

tests/devices/e1/device_e1_test.py
```diff
@@ -0,0 +1,81 @@
...
+from unittest.mock import patch
+
+import pytest
+
+from midealocal.const import ProtocolVersion
+from midealocal.devices.e1 import DeviceAttributes, MideaE1Device
+from midealocal.devices.e1.message import MessageQuery, MessageWork
+
+
+class TestMideaE1Device:
```
chemelli74, line 13: Can we please avoid the specific class and use instead a generic one in conftest.py so to reuse for all device types ?

chemelli74 (dismissed):

wuwentao:
## Review summary — `feat(e1): add dishwasher work controls`

Re-reviewed at `d8e9f78`, verified each finding with parallel checks, and pushed a fix for the one blocking convention issue (now head `5fd9f5e`). CI gates pass locally on the changed files: **247 passed**, `ruff check` ✅, `ruff format --check` ✅, `mypy` ✅.

### ✅ Fixed and pushed (`5fd9f5e`)
**Exception convention** — `set_work_mode`/`start_work` raised bare `ValueError`/`TypeError`. This class's own `set_attribute` already uses the project exception (`raise ValueWrongType` from `midealocal.exceptions`), so I routed all four raises through `ValueWrongType`, and updated the docstrings + the three E1 test assertions to match. This addresses @wuwentao's request to replace the raw `raise` with the library's exception hierarchy.

### 💬 Replies to the open review threads

**`__init__.py:150` — "Why copy()?" (@chemelli74):** It's intentional and correct — a defensive shallow copy so callers can't mutate the internal `_modes`, which `process_message` reads on the hot path (`self._modes.get(value)`). It's covered by `test_modes` (mutates the returned dict, asserts the original is unchanged). Recommend **keeping** it; a one-line `# defensive copy` comment would pre-empt the question.

**`__init__.py:173` — "Why not add a parameter with the work mode?" (@chemelli74):** `set_work_mode(mode)` and `start_work()` are *not* redundant — one is a stateless command (caller passes the code), the other starts the *already-selected* mode from device state. Adding a `mode` param to `start_work` would make it a duplicate of `set_work_mode`. **However**, the underlying concern is fair on two points that are worth addressing before merge:
- **Naming:** `set_work_mode`'s own docstring says it "Set… **and start** the selected program" — so *both* methods start the machine. The `set_`/`start_` verbs don't convey the real difference. Consider e.g. `start_work(mode)` + `start_selected_work()`.
- **`0x00` inconsistency:** `start_work` rejects `0x00`/"Neutral Gear" as "no mode selected", yet `set_work_mode(0x00)` accepts it. The two methods disagree on whether `0x00` is a valid mode — worth reconciling.

**`device_e1_test.py:12` — "use a generic fixture in conftest.py" (@chemelli74):** This E1 test uses the **same per-class fixture pattern** as the existing `da`/`e2`/`cd`/`a1`/`ac` device tests, and there is **no `tests/conftest.py`** in the repo today. So E1 is conforming to the current convention, not violating it — the request is really a repo-wide refactor (each device has different constructor args, so a shared fixture needs parametrization). Suggest tracking it as a **separate follow-up** that migrates all device tests together, rather than blocking this PR.

### Minor (non-blocking)
- `__init__.py` — the string `"[e1] No work mode selected"` is duplicated (two raise sites); could be a module constant.
- `start_work` reverse-resolves the stored mode *name* back to a code via string match, so a rename in `_modes` would silently break it; tracking the selected *code* would be more robust.

### Overall assessment: **request changes** (not yet ready to merge)
- ✅ **[FIXED]** `__init__.py` — bare `ValueError`/`TypeError` → `ValueWrongType` (pushed `5fd9f5e`)
- 🟡 **[DISCUSS — merge gate]** `__init__.py:156,173` — resolve the `set_`/`start_` naming and the `0x00`/"Neutral Gear" semantics inconsistency with the author (thread ③)
- 💬 **[ANSWERED]** `__init__.py:150` — `.copy()` is correct (defensive + tested); keep
- 🔵 **[SEPARATE PR]** `device_e1_test.py:12` — conftest fixture is a repo-wide follow-up, not a blocker here
- ⚪ **[SUGGESTION]** dedupe the repeated error string

Once the naming/`0x00` discussion (③) is settled and the `changes requested` review is re-approved, this should be good to merge — the exception-convention blocker is resolved and CI is green.

_Review processed by Claude Code (Opus 4.8)._

wuwentao:
@riccihenrique is there any review comments?  I just try to use claude code fixed some  review comments and commit/push to this PR, could you help to double check and confirm the result is ok for you ?

wuwentao (approved):

PR #476, opened 2026-06-22
fix: resolve TCP/UDP socket leaks and improve thread loop termination 

mrueg:
This PR addresses several issues regarding socket resource lifecycle management and thread termination loops in the codebase, preventing file descriptor leaks and potential application hangs/slowdowns when stopping device connection loops.

1. Robust Socket Shutdown & Close:
      • Wrapped  socket.shutdown(...)  inside  close_socket()  in device.py in a nested try-except block. This ensures  self._socket.close()  is always reached and executed, even if the socket was already disconnected by the peer (which causes  shutdown  to raise an  OSError ).
2. Explicit Socket Cleanup on Connection Error:
      • Modified all exception handlers inside  MideaDevice.connect()  to call  self.close_socket()  instead of merely dereferencing it or ignoring it.
3. Interruptible Reconnect Loop and Thread Termination:
      • Added a check for the thread run status ( self._is_run ) inside the connect loop  _connect_loop()  condition and during its sleep/exponential backoff delay. This allows thread reconnect cycles to terminate instantly rather than waiting for up to 10 minutes (600s).
      • Added a validation check for  self._is_run  before starting the query session in the main  run()  loop to prevent querying devices when the thread is stopped.
4. CLI Socket Leak Prevention:
      • Ensured device sockets are closed if they fail to authenticate or query attributes during discover in cli.py.
      • Wrapped CLI  set_attribute  command in a  try...finally  block to ensure all opened device sockets are properly closed before returning.
5. UDP Broadcast Socket Cleanup:
      • Wrapped UDP socket creation/usage in a  with socket.socket(...)  block inside discover.py to ensure it is automatically closed on return.

This should help with Midea's devices only supporting one TCP connection at a time.


midealocal/cli.py
```diff
@@ -131,10 +131,13 @@ async def discover(self) -> list[MideaDevice]:
                         dev.refresh_status(True)
                     except AuthException:
                         _LOGGER.debug("Unable to connect with key: %s", key)
+                        dev.close_socket()
                     except SocketException:
                         _LOGGER.exception("Device socket closed.")
+                        dev.close_socket()
                     except NoSupportedProtocol:
                         _LOGGER.exception("Unable to retrieve device attributes.")
+                        dev.close_socket()
```
chemelli74: Should be a finally

wuwentao:
@mrueg  Thank you very much for your PR and try to help us to imporve it.

we should double check with these changes in `device.py` loop part, as it may introduce some new issue, anyway, we can talk about the details.

1. there is some code comments in https://github.com/midea-lan/midea-local/blob/d9c1f14710d8949d3e8d79dff3462c24774c9818/midealocal/device.py#L689
 for example: HA is running without reboot for a long time, and user may power on/off midea devices(by mistaken or any reason), if device become offline for some hours and user power on the device again, device should become online, but HA may can't connect it again. 

2. device open() will only be call during HA setup entry via https://github.com/wuwentao/midea_ac_lan/blob/9f1b7dc20da0ea10b6b5a7b218d4a139922ef2cb/custom_components/midea_ac_lan/__init__.py#L245, if we set `_is_run` to `False` and add check it in loop , user only can reboot HA to recovery power off/on device, so `waiting for up to 10 minutes (600s)` is a solution to reduce HA resource and also support user devices still can become online and reconnect.

midealocal/cli.py
mrueg: changed, thanks!

wuwentao:
Thank for the update, I have no new review comments

mrueg:
@wuwentao thanks for the feedback and the review! Let me know if you need anything else to get it merged. :)

wuwentao:
thanks,I have no any new comments, and I just update it with latest main branch,  wait for ci result, also found there is a review comments from @chemelli74 about the `finally`? could you help to check and confirm the result or resove it?
as github show me 
```
Merging is blocked
All comments must be resolved.
```

anyway, we will delay some days to publish a new release, try to accept more PRs and feature/bug, then sync it to midea_ac_lan.
Milestone flag is ready, I will merge it before publish a new release.

midealocal/cli.py
```diff
@@ -229,22 +233,26 @@ async def download(self) -> None:
     async def set_attribute(self) -> None:
         """Set attribute for device."""
         device_list = await self.discover()
-        if len(device_list) != 1:
-            return
+        try:
+            if len(device_list) != 1:
+                return
```
chemelli74, line 243: Why a if should be under a try block ?

midealocal/device.py
```diff
@@ -223,24 +223,26 @@ def connect(self, check_protocol: bool = False) -> bool:
...
         except AuthException:  # authenticate exception
             _LOGGER.debug("[%s] Authentication failed", self._device_id)
+            self.close_socket()
         except SocketException:  # refresh_status exception
             _LOGGER.debug("[%s] Connect socket exception", self._device_id)
-            self._socket = None
+            self.close_socket()
         except NoSupportedProtocol:  # refresh_status exception
             _LOGGER.debug("[%s] No supported query protocol", self._device_id)
+            self.close_socket()
```
chemelli74: what about a finally ?

midealocal/device.py
```diff
@@ -619,7 +621,14 @@ def close_socket(self) -> None:
         self._buffer = b""
         if self._socket:
             try:
-                self._socket.shutdown(socket.SHUT_RDWR)
+                try:
```
chemelli74: Why nesting try block ?

midealocal/device.py
```diff
@@ -665,7 +674,10 @@ def _connect_loop(self) -> None:
                     sleep_time,
                 )
                 # sleep and reconnect loop until device online
-                time.sleep(sleep_time)
+                for _ in range(sleep_time):
+                    if not self._is_run:
+                        break  # type: ignore[unreachable]
```
chemelli74: ???

midealocal/device.py
```diff
@@ -697,6 +709,8 @@ def run(self) -> None:
         while self._is_run:
             # connect loop until device online
             self._connect_loop()
+            if not self._is_run:
+                break  # type: ignore[unreachable]
```
chemelli74: ???

tests/device_test.py
```diff
@@ -72,6 +72,10 @@ def test_connect(self, exc: Exception, result: bool) -> None:
         with patch("socket.socket.connect", side_effect=exc):
             assert self.device.connect() is result
             assert self.device.available is result
+            if not result:
+                assert self.device._socket is None
+            else:
+                assert self.device._socket is not None
```
chemelli74: Use parametrize, not if in tests please

chemelli74 (dismissed):

wuwentao:
## Code review — socket lifecycle and thread termination

Reviewed the current revision (head `af9cd61`). The core intent is sound: ensuring sockets are released on error paths and making the thread loop interruptible. Most of the mechanical changes (replacing `self._socket = None` with `self.close_socket()`, wrapping `shutdown()` in an inner try/except, wrapping discover.py's UDP socket in `with`) are correct improvements. A few gaps and one subtle regression worth addressing before merge.

### Correctness

1. **cli.py `discover()` — uncaught exceptions abort the whole scan and leak already-found devices' sockets** (lines 132, 136-141). `refresh_status()` and `authenticate()` can raise `OSError`/`TimeoutError`/`ConnectionResetError` (from bare `socket.send()`/`.recv()` calls in device.py, and `send_message_v2` explicitly re-raises them at lines 304-334). These are **not** in the three `except` clauses (`AuthException`, `SocketException`, `NoSupportedProtocol`). When such an exception is raised, the `finally` closes the *current* device's socket correctly, but the exception then propagates out of `discover()`, skipping `return device_list` (line 145) → all devices already appended to `device_list` (with `success=True` and live sockets) are lost and their sockets leaked. This directly undercuts the PR's cleanup goal. Suggest either catching `OSError`/`TimeoutError`/`ConnectionResetError` explicitly, or wrapping the finally in a broader outer try/finally that releases all devices in `device_list` on any escape (similar to `set_attribute()`).

2. **`_connect_loop` calls `connect()` after `stop()`** (device.py line 666). The while guard `self._socket is None and self._is_run` (line 664) exits when `stop()` sets `_is_run=False` (device never connected). But line 666 still calls `if self.connect(check_protocol=True) is False:`. `connect()` (lines 205-249) creates a socket and performs network operations (authenticate, refresh_status) without checking `_is_run`. If `stop()` was called, `connect()` proceeds anyway, creating a socket during shutdown. The `close_socket()` at line 667 cleans it up, but the unnecessary socket creation and round-trips happen after termination was requested. Suggest skipping the `connect()` call if the loop exited via `_is_run` (e.g. `if self._is_run and self.connect(...) is False:`).

3. **Double `authenticate()` for V3 devices** (cli.py line 130). `connect()` already calls `authenticate()` internally for V3 devices (device.py line 217-218). Then `discover()` calls `dev.authenticate()` a second time on the same socket, re-running the handshake. If the device rejects a repeat handshake (returns <20 bytes), `authenticate()` raises `AuthException` (device.py line 277), the except swallows it, and a genuinely working device is dropped from discovery. **Note:** This is pre-existing (these are unchanged context lines in the diff), not introduced by this PR — but it's a real defect in the code under review.

### Test coverage

4. **Test assertions don't verify the cleanup fix** (tests/device_test.py lines 235-238). The new assertions only check `_socket is None`, which would pass both before and after the PR (the old code set `self._socket = None` directly). They don't verify that `_buffer` was cleared, `_unsupported_protocol` was reset, or `shutdown()`/`close()` were called — so the test doesn't distinguish the old leak from the fix. Consider mocking `close_socket()` and asserting it was called, or asserting on `_buffer`/`_unsupported_protocol` state.

### Fragility / altitude

5. **`# type: ignore[unreachable]` suppressions** (device.py lines 679, 713). Mypy proves the breaks unreachable under single-thread analysis (after `while self._is_run`, it narrows `_is_run` to True). The code works because `_is_run` is mutated cross-thread (via `stop()`), which mypy can't see. The suppressions are load-bearing but fragile — a bare `[unreachable]` comment looks like acknowledged dead code to delete, inviting cleanup that breaks termination. A comment clarifying the cross-thread mutation would help, or adopting `threading.Event` (see #6) removes the need for the ignores entirely.

6. **Busy-poll sleep vs `threading.Event`** (device.py line 677). The interruptible sleep `for _ in range(sleep_time): if not self._is_run: break; time.sleep(1)` reimplements an interruptible wait with 1-second granularity. `threading.Event.wait(timeout)` is the idiomatic tool: `self._stop_event.wait(sleep_time)` returns immediately when set, has no granularity loss, and is opaque to mypy (so the fragile `[unreachable]` suppressions disappear). A device 590s into a 600s backoff blocks close()/shutdown for ~1 more second with the current approach. This is an altitude observation — consider it for a follow-up, not a blocker for this PR.

### Cleanup

7. **`connect()` repeats `close_socket()` in 6 handlers** (device.py lines 224-245). All 6 exception handlers call the identical `self.close_socket()`. Since `connected` is set True only on success, a single `finally: if not connected: self.close_socket()` would dedup them all. Notably, the PR already applies this exact success-flag + finally pattern in cli.py `discover()` (lines 126-144) but leaves the 6 repeated calls in `connect()` — an internal inconsistency. The finally form is more maintainable (a 7th exception type can't forget its cleanup).

---

**Highest-value fixes:** catch or wrap the uncaught socket exceptions in `discover()` (#1), skip `connect()` after `_is_run` goes False (#2), and strengthen the test assertions (#4). The double-auth (#3) is pre-existing and lower-risk but worth addressing.

---

*Review generated via [Claude Code](https://claude.ai/code) with max-effort recall mode (10 finder angles, verification, gap sweep). Model: Claude Opus 4.8.*

wuwentao (commented):
Follow-up to my summary comment above: concrete fix suggestions inline on the relevant lines. Four are one-click committable `suggestion` blocks (#1 uncaught exceptions, #2 connect-after-stop, the test assertions, and the `connect()` `finally` dedup); the double-auth and `threading.Event` items are code examples since they touch more than one spot.

_Generated via [Claude Code](https://claude.ai/code) (Opus 4.8)._

midealocal/cli.py
```diff
@@ -123,21 +123,25 @@ async def discover(self) -> list[MideaDevice]:
...
                         dev.refresh_status(True)
+                        _LOGGER.info("Found device:\n%s", dev.attributes)
+                        device_list.append(dev)
+                        success = True
                     except AuthException:
                         _LOGGER.debug("Unable to connect with key: %s", key)
                     except SocketException:
                         _LOGGER.exception("Device socket closed.")
                     except NoSupportedProtocol:
                         _LOGGER.exception("Unable to retrieve device attributes.")
```
wuwentao, line 141: **Uncaught socket errors abort the whole scan and leak already-found devices' sockets.**

`refresh_status()` / `authenticate()` can raise `OSError` (and its subclasses `TimeoutError`, `ConnectionResetError`) — `send_message_v2` re-raises them explicitly (device.py 304-334). None are in these three `except` clauses, so such an exception escapes `discover()` after the `finally` closes only the *current* device, skipping `return device_list` and leaking every device already appended.

Add a handler so one bad device doesn't abort discovery (the `finally` still closes it since `success` stays `False`). `OSError` covers `TimeoutError`/`ConnectionResetError`:

```suggestion
                    except AuthException:
                        _LOGGER.debug("Unable to connect with key: %s", key)
                    except SocketException:
                        _LOGGER.exception("Device socket closed.")
                    except NoSupportedProtocol:
                        _LOGGER.exception("Unable to retrieve device attributes.")
                    except OSError:
                        _LOGGER.exception("Connection error during device query.")
```

midealocal/device.py
```diff
@@ -652,7 +661,7 @@ def _connect_loop(self) -> None:
         """Connect loop until device online."""
         # connect loop until online
         connection_retries = 0
-        while self._socket is None:
+        while self._socket is None and self._is_run:
             _LOGGER.debug("[%s] Socket is None, try to connect", self._device_id)
             if self.connect(check_protocol=True) is False:
```
wuwentao: **`connect()` is still called after `stop()` requested shutdown.**

The `while` guard checks `_is_run`, but if `stop()` flips it to `False` during the previous iteration's `connect()`/sleep, control still reaches this line and opens a new socket + does network I/O during teardown. Short-circuit on `_is_run` so `connect()` is skipped once shutdown is requested:

```suggestion
            if self._is_run and self.connect(check_protocol=True) is False:
```

tests/device_test.py
wuwentao: **These assertions don't actually cover the fix.**

Asserting `_socket is None` passed *before* this PR too (the old code did `self._socket = None`). To lock in the new `close_socket()` behavior, also assert the state that only `close_socket()` resets:

```suggestion
            if not result:
                assert self.device._socket is None
                assert self.device._buffer == b""
                assert self.device._unsupported_protocol == []
            else:
                assert self.device._socket is not None
```

midealocal/device.py
```diff
@@ -223,24 +223,26 @@ def connect(self, check_protocol: bool = False) -> bool:
...
             _LOGGER.debug("[%s] No supported query protocol", self._device_id)
+            self.close_socket()
         except Exception as e:
             _LOGGER.exception(
                 "[%s] Unknown error during connect device",
                 self._device_id,
                 exc_info=e,
             )
-            self._socket = None
+            self.close_socket()
```
wuwentao: **The 6 identical `close_socket()` calls can collapse into one `finally`** — the same success-flag + finally idiom this PR introduces in `cli.py discover()`. This is leak-proof by construction: a 7th `except` added later can't forget to clean up.

```suggestion
        except TimeoutError:
            _LOGGER.debug("[%s] Connection timed out", self._device_id)
        except OSError:  # refresh_status exception
            _LOGGER.debug("[%s] Connection error", self._device_id)
        except AuthException:  # authenticate exception
            _LOGGER.debug("[%s] Authentication failed", self._device_id)
        except SocketException:  # refresh_status exception
            _LOGGER.debug("[%s] Connect socket exception", self._device_id)
        except NoSupportedProtocol:  # refresh_status exception
            _LOGGER.debug("[%s] No supported query protocol", self._device_id)
        except Exception as e:
            _LOGGER.exception(
                "[%s] Unknown error during connect device",
                self._device_id,
                exc_info=e,
            )
        finally:
            if not connected:
                self.close_socket()
```

midealocal/cli.py
```diff
@@ -123,21 +123,25 @@ async def discover(self) -> list[MideaDevice]:
                 )
                 _LOGGER.debug("Opening socket for device.")
                 if dev.connect():
+                    success = False
                     try:
                         if device["protocol"] == ProtocolVersion.V3:
                             _LOGGER.debug("Trying to connect with key: %s", key)
                             dev.authenticate()
```
wuwentao: **Redundant second authenticate (pre-existing, not introduced by this PR).**

`dev.connect()` above already calls `authenticate()` internally for V3 devices (device.py 217-218), so this is a second handshake on the same socket. A device that rejects a repeat handshake makes `authenticate()` raise `AuthException`, dropping a device that connected fine. Consider dropping the redundant call:

```python
                    try:
                        # connect() already authenticated V3 devices
                        _LOGGER.debug("Trying to retrieve device attributes.")
                        dev.refresh_status(True)
```

midealocal/device.py
```diff
@@ -665,7 +674,10 @@ def _connect_loop(self) -> None:
                     sleep_time,
                 )
                 # sleep and reconnect loop until device online
-                time.sleep(sleep_time)
+                for _ in range(sleep_time):
+                    if not self._is_run:
+                        break  # type: ignore[unreachable]
+                    time.sleep(1)
```
wuwentao, line 721: **Optional / follow-up (altitude).** This busy-poll wait with 1s granularity plus the `# type: ignore[unreachable]` on line 679 (and 713) both exist because `_is_run` is a plain bool that mypy can't see mutated cross-thread. A `threading.Event` is the idiomatic primitive and removes both the loop and the fragile ignores:

```python
    # in __init__:
    self._stop_event = threading.Event()
    # stop(): self._stop_event.set()
    # start(): self._stop_event.clear()

    # replace the for/sleep loop with a single interruptible wait:
                if self._stop_event.wait(sleep_time):
                    break  # woken by stop()
```
`Event.wait(timeout)` returns immediately when set (no 1s latency) and is opaque to mypy, so the `[unreachable]` suppressions on lines 679/713 disappear.

wuwentao:
## Review fixes applied — `socket-closure`

Pushed `9d9d8ee` addressing the open review comments from @chemelli74 and @wuwentao. Verified locally: **242 tests pass**, `ruff check` + `ruff format --check` clean (with the CI-pinned ruff `v0.6.9`), and `mypy` clean.

### Changes

| Comment | Resolution |
|---------|-----------|
| **@chemelli74** `device.py:238` *"what about a finally?"* / **@wuwentao** `device.py:245` (6× close_socket) | `connect()` now releases the socket once in `finally: if not connected: close_socket()` instead of repeating the call in all six `except` handlers — leak-proof for any future exception path too. |
| **@chemelli74** `device.py:624` *"Why nesting try?"* | Flattened `close_socket()`'s nested try/except — `shutdown()` and `close()` each get their own guard, no nesting. |
| **@chemelli74** `device.py:679` & `:713` *"???"* / **@wuwentao** `:680` (type:ignore) | Replaced both `# type: ignore[unreachable]` with a small `_should_run()` accessor. `_is_run` is flipped from another thread; reading it through a method keeps the loop-exit `break` reachable to mypy — no suppression needed (mypy confirmed clean). |
| **@chemelli74** `tests/device_test.py:78` *"Use parametrize, not if in tests"* / **@wuwentao** `:78` (assertions) | `test_connect` now parametrizes the socket-state expectation (`socket_is_none`) instead of `if/else`, and asserts `close_socket()`'s **full** effect (buffer cleared) on failure paths — so it actually covers the new behaviour, not the pre-existing `_socket = None`. |
| **@wuwentao** `cli.py:141` (uncaught exceptions abort scan) | `discover()` now also catches `OSError` (covers `TimeoutError`/`ConnectionResetError` from `authenticate()`/`refresh_status()`), so one unreachable device can't abort the whole scan and leak already-found devices' sockets. |

### Notes on two comments left as-is (intentionally)

- **@chemelli74** `cli.py:238` *"Why a `if` under a try block?"* — the `if len(device_list) != 1: return` is inside the `try` **on purpose**: the `finally` must still close every socket from `discover()` even on that early return. Moving the guard before the `try` would leak those sockets. Kept as-is; happy to add a clarifying comment if preferred.
- **@wuwentao** `cli.py:130` (redundant V3 `authenticate()`) — this is **pre-existing** (unchanged context lines, not introduced by this PR), so I've left it out of this change and will open a **separate PR** for it to keep this one scoped to the socket-closure review.

### Status
CI should re-run on `9d9d8ee`. Leaving the final review/merge decision to @chemelli74 (who has `changes requested`) — I've addressed the raised points but am not self-approving a maintainer-reviewed PR. Once the CI rerun is green and @chemelli74 is happy, it should be ready to merge.

_Fixes processed by Claude Code (Opus 4.8)._

wuwentao:
@chemelli74  @mrueg 
could you help to review and confirm the current code?

wuwentao (approved):

chemelli74:
If you don't mind I would like to review this carefully as I had a hint from HomeAssistant core on how to fix threading

wuwentao:
## Update — merged `main`, resolved the conflict, and applied the one remaining fix

Pushed `bcbb0c9` to `socket-closure`. The branch was `CONFLICTING`/`DIRTY` against `main`; it is now **MERGEABLE**. Net PR diff is unchanged in scope (4 files: `cli.py`, `device.py`, `discover.py`, `device_test.py`).

### What changed in this push
1. **Merged current `main`** into the branch (it was 22 commits behind).
2. **Resolved the `tests/device_test.py` conflict** in `test_connect` — combined `main`'s `patch.object(authenticate/refresh_status)` + `connect(check_protocol=True)` (from #514 "restore silently skipped tests") with this PR's `socket_is_none` parametrize column and the `_buffer == b""` failure-path assertion. Both intents are preserved.
3. **Applied the one open correctness fix** — `connect()` after shutdown (`device.py`, `_connect_loop`):
   ```python
   if self._should_run() and self.connect(check_protocol=True) is False:
   ```
   `_should_run()` is re-checked before `connect()` so no socket/network I/O starts once `close()` has flipped `_is_run` (the `while` guard alone can't cover a shutdown that lands between the guard and the call). Uses the existing `_should_run()` helper for consistency and to stay visible to mypy.

### Status of prior review findings
| Finding | Status |
|---|---|
| Uncaught `OSError` aborts scan + leaks sockets (`cli.py`) | ✅ fixed (contributor added `except OSError`) |
| `close_socket()` nested try → sequential try blocks | ✅ fixed |
| Fragile `# type: ignore[unreachable]` | ✅ fixed (`_should_run()` helper) |
| Test assertions didn't cover the fix | ✅ fixed (parametrized + `_buffer`) |
| `connect()` after `stop()` | ✅ **fixed in this push** |
| 6× `close_socket()` → single `finally` in `connect()` | ✅ fixed |
| `if len(device_list) != 1: return` inside `try` | ℹ️ intentional — replied in thread (kept inside `try` so `finally` closes sockets on early return) |
| Redundant V3 double-auth in `discover()` | ↪️ pre-existing, **deferred to a follow-up issue** (out of scope for socket-closure) |
| `threading.Event` instead of busy-poll sleep | ↪️ optional altitude, **deferred to a follow-up** |

### Verification (local, on `bcbb0c9`)
- `ruff check` (pinned `v0.6.9` rules) — clean on all 4 changed files
- `mypy midealocal/device.py cli.py discover.py` — Success, no issues
- `pytest tests/` — **270 passed** (incl. all 6 `test_connect` cases)

@chemelli74 — could you take another look and resolve the threads if this addresses your comments? The two deferred items (double-auth, `threading.Event`) I'd suggest as separate follow-up issues rather than expanding this PR's scope.

---
*Reviewed & fixed via [Claude Code](https://claude.com/claude-code) — model: Claude Opus 4.8. 🤖*

midealocal/cli.py
wuwentao: The early `return` needs to be **inside** the `try` so the `finally` still runs on that path and closes every socket.

`discover()` returns devices whose sockets are **left open** on success (it only calls `close_socket()` on the failure branch, so callers can talk to the device). So when `discover()` finds ≠ 1 device — e.g. 2 devices — `set_attribute()` must still close all of them before returning. If the `if len(device_list) != 1: return` were moved above the `try`, that early return would bypass `finally` and leak those open sockets. Keeping it inside is what guarantees cleanup on both the early-return and the normal/exception paths.

If the plain `if/return` inside `try` reads oddly, an equivalent alternative is to close-and-return before the `try`:

```python
device_list = await self.discover()
if len(device_list) != 1:
    for dev in device_list:
        dev.close_socket()
    return

try:
    ...  # set_attribute / refresh_status
finally:
    device_list[0].close_socket()
```

Both are correct; the current form just avoids duplicating the close loop. Happy to switch to whichever you prefer.

chemelli74:
@wuwentao, I kindly asked to wait as I wanted to do a in depth review.
May I know why you merged it ?
Thank you

wuwentao:
sorry, all the PRs waiting for your review and you are too busy to reply or post review comments.
and I also post more info and requires in another PR  #493

if you really have review comments and want to improve it. you still can post new PR to finish it.
and NOT always request all of us to wait for you with more weeks.

please check and confirm PRs status and review comments, We have a responsibility and obligation to respond to the PR owner regarding our review comments as soon as possible, rather than keeping everyone waiting for us to find the time to address them.

chemelli74:
Haste isn't always a good thing ;-)

For example, look at the PR to add Midea LAN to the HomeAssistant core: it took months to merge, but now the code is clean and maintainable; in addition, we fixed a few bugs that hadn't been inspected in the custom component.

wuwentao:
> Haste isn't always a good thing ;-)
> 
> For example, look at the PR to add Midea LAN to the HomeAssistant core: it took months to merge, but now the code is clean and maintainable; in addition, we fixed a few bugs that hadn't been inspected in the custom component.

I think you should weigh the costs and benefits. While focusing on continuously refining a specific feature can certainly yield high-quality results, rejecting dozens of pull requests (PRs) of slightly lower quality simply due to a lack of time risks driving those contributors away entirely; they will be far less likely to engage or submit PRs in the future. The impact and cost of losing that community engagement far outweigh the benefits of obsessively perfecting a single feature.
Finally, this repository—in terms of code quality, user base, scope of use, and foreseeable lifecycle—does not yet match the scale or influence of major projects like HA Core or HACS. (Domestically, Midea is continuously iterating on new products and phasing out the old protocol that was reverse-engineered on GitHub; there are already several repositories for newer Midea products offering similar functionality.) We need to have a clear understanding of where this project stands and acknowledge the gap between it and large-scale projects; it is unreasonable to apply the standards of massive projects to a modest one, as the costs and trade-offs involved are fundamentally different.

PR #478, opened 2026-06-23
feat(ac): expose per-mode min/max temperature from B5 capability

Pulpyyyy:
## Summary
AC climate temperature bounds are hardcoded to 16–30 °C in the downstream HA integration. This PR makes the real bounds available from the device itself.

The B5 capability already carries a temperature node (tag `0x0225`) with per-mode limits, but it was parsed (`b5_temperature0..6`) and never used. This PR turns it into two new attributes:
- **`min_temperature`** / **`max_temperature`** — resolved against the current mode.

## Behaviour
- Layout `[cool_min, cool_max, auto_min, auto_max, heat_min, heat_max, flag]` (0.5 °C units).
- Mode mapping: 1 auto, 2 cool, 3 dry, 4 heat, 5 fan. dry/fan and any unknown/off mode reuse the **cool** range.
- **No regression**: defaults stay `16.0` / `30.0` when the device does not report the capability.

## Verification
Tested live against a real portable AC (model `00000Q17`) whose B5 declares `22 3c 22 3c 22 3c 00` → **17–30 °C**. Before, HA showed `min_temp: 16`; the device now reports `min_temperature: 17.0`, `max_temperature: 30.0`.

## CI
`mypy` (strict) passes locally; lines ≤ 88; dynamic attribute access is guarded by a single `hasattr` (same pattern as `self_clean_active`).

_Required by wuwentao/midea_ac_lan#857, which consumes these bounds for the climate min/max temperature and documents the new `min_temperature` / `max_temperature` customize options._

### DIFF
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -37,6 +37,8 @@ class DeviceAttributes(StrEnum):
     power = "power"
     mode = "mode"
     target_temperature = "target_temperature"
+    min_temperature = "min_temperature"
+    max_temperature = "max_temperature"
     fan_speed = "fan_speed"
     swing_vertical = "swing_vertical"
     swing_horizontal = "swing_horizontal"
@@ -140,6 +142,8 @@ class MideaACDevice(MideaDevice):
                 DeviceAttributes.power: False,
                 DeviceAttributes.mode: 0,
                 DeviceAttributes.target_temperature: 24.0,
+                DeviceAttributes.min_temperature: 16.0,
+                DeviceAttributes.max_temperature: 30.0,
                 DeviceAttributes.fan_speed: 102,
                 DeviceAttributes.swing_vertical: False,
                 DeviceAttributes.swing_horizontal: False,
@@ -189,6 +193,8 @@ class MideaACDevice(MideaDevice):
         self._used_subprotocol: bool = False
         self._bb_sn8_flag: bool = False
         self._bb_timer: bool = False
+        # per-mode setpoint limits from the B5 capability, keyed by mode value
+        self._temperature_limits: dict[int, tuple[float, float]] | None = None
         self._power_analysis_method: int = 1
         self._default_power_analysis_method: int = 1
         self.set_customize(customize)
@@ -296,7 +302,29 @@ class MideaACDevice(MideaDevice):
             active = message.self_clean_active
             self._attributes[DeviceAttributes.self_clean] = active
             new_status[DeviceAttributes.self_clean.value] = active
+        new_status.update(self._update_temperature_limits(message))
         return new_status
+
+    def _update_temperature_limits(self, message: MessageACResponse) -> dict[str, Any]:
+        """Update the per-mode setpoint limits from the B5 capability.
+
+        An unknown mode (e.g. 0 when off) falls back to the cool range.
+        """
+        if hasattr(message, "temperature_limits"):
+            self._temperature_limits = message.temperature_limits
+        if self._temperature_limits is None:
+            return {}
+        mode = self._attributes[DeviceAttributes.mode]
+        minimum, maximum = self._temperature_limits.get(
+            mode,
+            self._temperature_limits[2],
+        )
+        self._attributes[DeviceAttributes.min_temperature] = minimum
+        self._attributes[DeviceAttributes.max_temperature] = maximum
+        return {
+            DeviceAttributes.min_temperature.value: minimum,
+            DeviceAttributes.max_temperature.value: maximum,
+        }
 
     def make_message_set(self) -> MessageGeneralSet:
         """Midea AC device make message set."""
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -142,8 +142,8 @@ class MideaACDevice(MideaDevice):
                 DeviceAttributes.power: False,
                 DeviceAttributes.mode: 0,
                 DeviceAttributes.target_temperature: 24.0,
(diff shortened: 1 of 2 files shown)


wuwentao:
@Pulpyyyy thank you very much for your PR.

I have an additional requirements: 
as this min/max temperature only works for B5 message, we can add a custommize option in `def _update_temperature_limits` ( like `temperature_step` option), refer to https://github.com/midea-lan/midea-local/blob/d9c1f14710d8949d3e8d79dff3462c24774c9818/midealocal/devices/ac/__init__.py#L525

1. custommize option have first priority, if result is None, continue check b5 message value like your current PR `hasattr(message, "temperature_limits")`, if both is None, refer to midea_ac_lan default const value.
2. once it done, all the AC devices should support temperature range manual set it or auto set it for b5 message, then we can add this custommize option to midea_ac_lan 0xAC device docs.

as some old device may support both min 16 or min 17, some github issue also required to support different min/max value with different device type(user should confirm the expected min/max value via Midea official app or remote control).

Pulpyyyy:
Thanks @wuwentao, done in 50a19ed.

Implemented the customize option with the exact priority you described:

1. **customize** (`min_temperature` / `max_temperature`) — highest priority
2. **B5 message** (`temperature_limits`) — when customize is None
3. **None** — when both are None, so the consumer (midea_ac_lan) falls back to its own default const

`min` and `max` resolve independently, so you can override just one (e.g. force `min_temperature: 16` while keeping the B5 `max`). The default attribute value is now `None` (instead of 16/30) so the const truly lives in midea_ac_lan.

Verified live on a portable AC (model 00000Q17):
- no customize → `17 / 30` (from B5)
- `{"min_temperature": 16}` → `16 / 30` (override min, B5 max)

mypy (strict) passes. I'll add the `min_temperature` / `max_temperature` customize option to the 0xAC docs in the midea_ac_lan PR https://github.com/wuwentao/midea_ac_lan/pull/857

wuwentao:
Thanks, LGTM
merged

Pulpyyyy:
Hi @wuwentao, a follow-up to the customize work here — **auto-deriving AC capabilities from the B5 message**.

While testing on a cooling-only portable AC (model `00000Q17`), I found the B5 capability bytes already describe the device's real feature set — they just need proper per-capability decoding. The raw byte is **not** a simple 0/1; each capability has its own value semantics (matching the `msmart` project).

Decoded against live hardware, it reconciles **9/9** with what the unit actually accepts/rejects:

| capability | id | raw | decoded |
| --- | --- | --- | --- |
| modes | 0x0214 | 5 | cool + dry + fan, **no heat/auto** |
| swing | 0x0215 | 2 | **none** (vertical & horizontal both off) |
| fan_speed | 0x0210 | 4 | low / high / auto |
| eco | 0x0212 | 0 | not supported |
| anion | 0x021E | 0 | not supported |
| turbo | 0x021A | 2 | not supported (cool & heat) |
| display | 0x0224 | 1 | supported |
| temperature | 0x0225 | 22 3c ... | 17–30 °C |

Proposal, mirroring the min/max temperature pattern from this PR: decode these in `XB5MessageBody`, expose capability flags from the device, and have `midea_ac_lan` derive `hvac_modes` / `fan_modes` / `supported_features` from them — with **customize override** (customize > B5 > default), exactly like the temperature range.

Would you welcome this in midealocal (+ midea_ac_lan)? And scope-wise: start with the climate entity (modes / fan / swing), with switch-entity filtering as a later step?

I already have a working decode (verified on hardware) ready to turn into a PR if you're on board.

wuwentao:
@Pulpyyyy thanks for the update, do you mean the `MessageNewProtocolQuery` with `NewProtocolTags` and parse response with `XB5MessageBody`, set command with `MessageNewProtocolSet` in [devices/ac/message.py](https://github.com/midea-lan/midea-local/blob/main/midealocal/devices/ac/message.py) ?
`NewProtocolTags`  should be the capability and id value.
you can download your device lua file and check the source code via lua script.
lua script related command can be found in `cli.py`

I just add some via old github issue, but I don't have any b5 device to test it, it's hard for me to do 0xAC device code chanages, and may introduce more issues with exist old devices, so only match the basic requirements.
anyway, you can try to improve it with new feature and test it with your b5 device, confirm the result.
new PR is welcome, midea-local lib also is ok, and midea_ac_lan to enable it, if it may introduce new issue, we can continue improve it and fix bug in future release.
thanks for your help and support.

Pulpyyyy:
Thanks @wuwentao! Yes, exactly — but **read-only**, so it's low-risk:

- The B5 response is already fetched (`MessageCapabilitiesQuery` in `build_query`) and parsed by `XB5MessageBody`. I only add the **interpretation** of each capability's value (the raw byte isn't 0/1 — each tag has its own value semantics, matching the `msmart` predicates), exposed as `MideaACDevice.capabilities`.
- **No `MessageNewProtocolSet` change needed** — we don't *set* anything new; the climate just reads `capabilities` to expose only the modes/fan/swing/presets the device actually supports.

To address your worry about 0xAC regressions: it's **purely additive** — a new `capabilities` dict populated only from B5, and the integration reads it via `getattr(..., {})`. Devices without B5 (or that don't report a capability) keep the current behaviour; priority is customize > B5 > default.

I downloaded the device Lua and decoded the values (verified 9/9 on a real portable AC `00000Q17`): https://gist.github.com/Pulpyyyy/02d3d42d62ab17c10ab3f799b00fdf61

Opened the two PRs:
- midea-local (decode + expose `capabilities`): #481
- midea_ac_lan (consume it): wuwentao/midea_ac_lan#862

Happy to iterate. Thanks!

PR #488, opened 2026-06-30
feat: report MAC address

zeehio:
This PR extracts the MAC address from the discover process and provides the attribute `.mac` to the device class, so https://github.com/wuwentao/midea_ac_lan/pull/868 can work

### DIFF
--- a/midealocal/device.py
+++ b/midealocal/device.py
@@ -6,7 +6,7 @@ import threading
 import time
 from collections.abc import Callable
 from enum import IntEnum, StrEnum
-from typing import Any, ClassVar, TypedDict, Unpack
+from typing import Any, ClassVar, NotRequired, TypedDict, Unpack
 
 from typing_extensions import deprecated
 
@@ -83,6 +83,7 @@ class MideaDeviceInitKwargs(TypedDict):
     device_protocol: ProtocolVersion
     model: str
     subtype: int
+    mac: NotRequired[str | None]
 
 
 class MideaDevice(threading.Thread):
@@ -123,6 +124,7 @@ class MideaDevice(threading.Thread):
         self._previous_refresh = 0.0
         self._previous_heartbeat = 0.0
         self.name = self._device_name
+        self.set_mac(kwargs.get("mac"))
 
     _fahrenheit_default: ClassVar[bool] = False
 
@@ -197,6 +199,11 @@ class MideaDevice(threading.Thread):
     def subtype(self) -> int:
         """Device subtype."""
         return self._subtype
+
+    @property
+    def mac(self) -> str | None:
+        """Device MAC address."""
+        return self._mac
 
     @staticmethod
     def fetch_v2_message(msg: bytes) -> tuple[list, bytes]:
@@ -647,6 +654,10 @@ class MideaDevice(threading.Thread):
             _LOGGER.debug("[%s] Update IP address to %s", self._device_id, ip_address)
             self._ip_address = ip_address
             self.close_socket()
+
+    def set_mac(self, mac: str | None) -> None:
+        """Set MAC."""
+        self._mac = mac or None
 
     def set_refresh_interval(self, refresh_interval: int) -> None:
         """Set refresh interval."""
--- a/midealocal/discover.py
+++ b/midealocal/discover.py
@@ -1,6 +1,7 @@
 """Midea local discover."""
 
 import logging
+import re
 import socket
 from ipaddress import IPv4Network
 from typing import Any
@@ -159,6 +160,35 @@ SERIAL_TYPE2_LENGTH = 22
 DISCOVERY_MIN_REPLY_LENGTH = 41
 
 
+def _extract_mac(reply: bytes | bytearray, ssid_len: int, sn: str) -> str | None:
+    """Return the MAC address or None if not available.
+
+    Extracts the MAC address from the v3 discover reply and returns it as a
+    hexadecimal string without internal separators.
+    If the reply does not provide the MAC address, the MAC is extracted from
+    the serial number, which also may contain the MAC address.
+    Otherwise return None.
+    """
+    # Based on https://github.com/nbogojevic/midea-beautiful-air/blob/db3622e784891af0a522d70a626fb54e5c3e5e6f/midea_beautiful/lan.py#L264  # noqa: E501
+    mac = None
+    mac_start = 63 + ssid_len
+    mac_ends = mac_start + 6
+    if len(reply) >= mac_ends:
+        mac = reply[mac_start:mac_ends].hex()
+    elif len(sn) == SERIAL_TYPE1_LENGTH:
+        mac = sn[16:28]
+    elif len(sn) == SERIAL_TYPE2_LENGTH:
+        # MAC address not reported
+        mac = None
+    else:
+        _LOGGER.warning("Unknown serial number format %s", sn)
+        mac = None
+    if mac and not re.fullmatch(r"[0-9A-Fa-f]{12}", mac):
+        _LOGGER.warning("Invalid MAC address %s", mac)
+        mac = None
+    return mac
+
+
 def _parse_discover_response(
     sock: socket.socket,
     found_devices: dict[int, dict[str, Any]],
@@ -190,13 +220,15 @@ def _parse_discover_response(
             40
         ] + DISCOVERY_MIN_REPLY_LENGTH <= len(reply):
             _LOGGER.debug("Declassified reply: %s", reply.hex())
+            ssid_len = reply[40]
             ssid = reply[
-                DISCOVERY_MIN_REPLY_LENGTH : DISCOVERY_MIN_REPLY_LENGTH + reply[40]
-            ].decode("utf-8")
+                DISCOVERY_MIN_REPLY_LENGTH : DISCOVERY_MIN_REPLY_LENGTH + ssid_len
+            ].decode("ascii")
             device_type = ssid.split("_")[1]
             port = bytes2port(reply[4:8])
-            model = reply[17:25].decode("utf-8")
-            sn = reply[8:40].decode("utf-8")
+            model = reply[17:25].decode("ascii")
+            sn = reply[8:40].decode("ascii")
+            mac = _extract_mac(reply, ssid_len, sn)
         else:
             _LOGGER.warning(
                 "Failed to decrypt the encrypt_data: %s from %s. Maybe not support.",
@@ -226,6 +258,7 @@ def _parse_discover_response(
             model = sn[3:11]
         else:
             model = ""
+        mac = _extract_mac(b"", 0, sn)
     else:
         return 0, None
     return device_id, {
@@ -236,6 +269,7 @@ def _parse_discover_response(
         "model": model,
         "sn": sn,
         "protocol": protocol,
+        "mac": mac,
     }
 
 
--- a/midealocal/cli.py
+++ b/midealocal/cli.py
@@ -120,6 +120,7 @@ class MideaCLI:
                     model=device["model"],
                     subtype=0,
                     customize="",
+                    mac=device["mac"],
                 )
                 _LOGGER.debug("Opening socket for device.")
                 if dev.connect():
--- a/midealocal/devices/__init__.py
+++ b/midealocal/devices/__init__.py
@@ -19,6 +19,7 @@ def device_selector(
     model: str,
     subtype: int,
     customize: str,
+    mac: str | None = None,
 ) -> MideaDevice:
     """Select and load device."""
     try:
@@ -38,6 +39,7 @@ def device_selector(
             model=model,
             subtype=subtype,
             customize=customize,
+            mac=mac,
         )
     except ModuleNotFoundError:
         device = None
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -16,6 +16,7 @@ from midealocal.cli import (
 from midealocal.cloud import MideaAirCloud, SmartHomeCloud
 from midealocal.const import ProtocolVersion
 from midealocal.device import AuthException, NoSupportedProtocol
+from midealocal.discover import _extract_mac
 from midealocal.exceptions import SocketException
 
 
@@ -102,6 +103,24 @@ class TestMideaCLI(IsolatedAsyncioTestCase):
             mock_default_keys.assert_called_once()
             mock_cloud_keys.assert_not_called()
 
+    def test_extract_mac(self) -> None:
+        """Test _extract_mac."""
+        expected_mac = "1234567890ab"
+        mac = _extract_mac(
+            reply=b"",
+            ssid_len=0,
+            sn="a" * 16 + expected_mac + "1234",
+        )
+        assert mac == expected_mac
+        mac2 = _extract_mac(
+            reply=b"a" * 63 + b"bb" + b"\x12\x34\x56\x78\x90\xab",
+            ssid_len=2,
+            sn="a" * 16 + expected_mac + "1234",
+        )
+        assert mac2 == expected_mac
+        mac3 = _extract_mac(b"", 1, "shortsn")
+        assert mac3 is None
+
     async def test_discover(self) -> None:
         """Test discover."""
         mock_device = {
@@ -111,7 +130,8 @@ class TestMideaCLI(IsolatedAsyncioTestCase):
             "ip_address": "192.168.0.2",
             "port": 6444,
             "model": "AC123000",
-            "sn": "0000AC12300000000000000000000000",
+            "sn": "0000AC12300000001234567890ABCDEF",
+            "mac": "1234567890AB",
         }
         mock_cloud_instance = AsyncMock()
         mock_device_instance = MagicMock()
--- a/tests/device_test.py
+++ b/tests/device_test.py
@@ -46,6 +46,7 @@ class MideaDeviceTest:
             model="test_model",
             subtype=1,
             attributes={},
+            mac="1234567890ab",
         )
 
     def test_initial_attributes(self) -> None:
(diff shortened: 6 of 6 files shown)


wuwentao:
thanks for your PR and support.
1. there is a doc for our PR and github action rule, please check it via https://github.com/midea-lan/midea-local?tab=contributing-ov-file
2. we need toupdate the PR title
3. the same rule as git commit history, github action will check it, please rebase all the history to one commit, as action will check it one by one, if anyone not match the rule, it will return failed and we can't merge it.

I just use my local `claude code` did a code review with this remote github PR url, and it give a brief summary as below:
```
 Summary

  The PR's mechanical plumbing is sound — all 40 device subclasses follow the existing model/subtype pattern exactly, _mac is always initialized (default None), the mac variable is bound on every code path
  in _parse_discover_response, and the ssid_len refactor is byte-identical to the prior inline reply[40].

  The substantive issues cluster in two places:

  - _extract_mac (the new logic) — a genuine correctness bug (#2: the SN-substring fallback returns something that is not a MAC, in a different format from the primary branch), a fragile guard (#3), plus
  it's the source of four build-breaking ruff violations (#4–#8) because it was written without the annotations/formatting the rest of this select=ALL codebase requires.
  - End-to-end wiring (#1) — the headline gap: discovery extracts the MAC, but the in-repo CLI never forwards it, so within this repo the "Report MAC address" feature is inert. (It does enable the external
  HA integration to pass mac, which is presumably the real intent — worth confirming with the author that no in-repo consumer was meant to be updated.)

  Highest-value fixes: wire mac=device["mac"] through the cli.py caller (#1, and update the mock_device test fixture so it doesn't KeyError), reconcile the two return formats in _extract_mac (#2), and add
  the missing type annotations + run ruff --fix to clear #4–#8 before CI does it for you.

  A note on confidence: I verified the lint findings by direct byte inspection (the rules are mechanical and ruff.toml confirms they're enabled), but I could not execute ruff here — ruff isn't installed and
  the sandbox couldn't reach PyPI to install it. If you want a definitive lint pass, run ruff check midealocal/discover.py midealocal/device.py locally.
```

zeehio:
Thank you very much for your review.

I will follow all your feedback. It may take me a while. I will leave the PR in draft until it's ready for another review

zeehio:
I don't have the vscode+devcontainers right now, but the CI is now passing on my fork so I guess it's ready for another review

wuwentao:
## Code review — MAC address reporting

Reviewed the current revision (head `69ea952`). The plumbing through `device_selector` and the ~40 device `__init__`s looks clean and consistent with the existing `model`/`subtype` pattern. A few issues in the new `_extract_mac` / `set_mac` logic that are worth addressing before merge, most severe first.

### Correctness

1. **`_extract_mac` fallback returns a 16-char value that isn't a valid MAC** (`discover.py`). The primary branch returns `reply[63+ssid_len:69+ssid_len].hex()` = 6 bytes → **12** hex chars, but the fallback returns `sn[16:32]` = a **16**-char substring of the serial. The new regex `re.fullmatch(r"[0-9A-Fa-f]+", mac)` validates *hex-ness* but not *length*, so the 16-char value passes and is reported as the MAC. The test even hardcodes `"mac": "1234567890ABCDEF"` (16 chars), locking in the malformed value. Suggest also enforcing `len(mac) == 12`.

2. **Protocol-1 (XML) path hardcodes `mac = None`** and never derives a MAC, even when its serial is 32 chars — inconsistent with the protocol-2/3 fallback now that every discovered device carries a `mac`.

3. **`len(sn) >= SERIAL_TYPE1_LENGTH` is checked on the decoded string**, but `sn = reply[8:40].decode("utf-8")` can be shorter than 32 chars if those 32 bytes contain any multibyte UTF-8, so the fallback is silently skipped. The check likely wants byte length (`len(reply[8:40])`). Also `sn is not None` is always true here (dead check).

4. **The hex regex is effectively dead for the primary branch** (`.hex()` is always lowercase hex) and only gates the SN fallback, where it nulls the MAC whenever the substring has a non-hex char — so SN-derived reporting becomes charset-dependent (a value for some serials, `None` for others).

5. **`set_mac("")` and `__init__(mac="")` disagree**: the constructor stores `""` verbatim, while `set_mac` normalizes `""` → `None`. Same input, different state.

### Cleanup / CI

6. **`set_mac` appears to be dead code** — no caller in the repo (MAC is set via the constructor kwarg). If it's intended for external consumers, worth a test; otherwise consider dropping it. Its body also simplifies to `self._mac = mac or None`.

7. **Lint (ruff `select = ["ALL"]`): the 138-char attribution comment in `_extract_mac` exceeds the line-length limit** and has no `# noqa: E501` — this will fail the lint job. Minor extras in the same function: redundant `bool()` inside `not bool(re.fullmatch(...))`, bare `63`/`69` offsets vs. the file's named-constant style, and a docstring typo "MAX address" → "MAC address".

### Testing

8. **The MAC extraction is never actually exercised** — `test_discover` mocks `discover`, so `_extract_mac`, the regex, and the reply/SN slicing don't run; only the mock dict's literal value is consumed (and the new SN doesn't match the fixture's `model`). A direct unit test feeding a real reply/SN through `_extract_mac` would catch the length bug in #1 and guard the offsets.

Happy to help with any of these if useful.

midealocal/discover.py
```diff
@@ -159,6 +160,27 @@
...
+    If the reply does not provide the MAX address, the MAC is extracted from
+    the serial number, which also may contain the MAC address.
+    Otherwise return None.
+    """
+    # Based on https://github.com/nbogojevic/midea-beautiful-air/blob/db3622e784891af0a522d70a626fb54e5c3e5e6f/midea_beautiful/lan.py#L264
+    mac = None
+    if len(reply) >= (69 + ssid_len):
+        mac = reply[63 + ssid_len : 69 + ssid_len].hex()
+    elif sn is not None and len(sn) >= SERIAL_TYPE1_LENGTH:
+        mac = sn[16:32]
```
wuwentao: **Medium - `mac` format/length is inconsistent across the two extraction sources.**

The reply path (line 175) returns `reply[63 + ssid_len : 69 + ssid_len].hex()` -> a real 6-byte MAC as **12 lowercase** hex chars. This serial fallback returns `sn[16:32]` -> **16 chars** taken verbatim from the serial (8 bytes, original case). So `device["mac"]` is sometimes a 12-char lowercase MAC and sometimes a 16-char uppercase serial fragment.

Any consumer treating this as a real MAC (e.g. Home Assistant `format_mac`, a unique-id / `connections` value, or splitting into 6 octets) will break for the fallback case. Consider normalizing both paths to one canonical form (lowercase, and only accept a true 12-hex-char MAC otherwise `None`) and updating the `tests/cli_test.py` expectation accordingly.

midealocal/discover.py
```diff
@@ -159,6 +160,27 @@
 DISCOVERY_MIN_REPLY_LENGTH = 41
 
 
+def _extract_mac(reply: bytes | bytearray, ssid_len: int, sn: str) -> str | None:
```
wuwentao, line 163: **Low - no direct test coverage for `_extract_mac`.**

The three behaviors here - reply-offset extraction (line 175), serial fallback (line 177), and the invalid->`None` regex guard (lines 178-180) - aren't covered by any `discover` unit test; `tests/cli_test.py` mocks `discover` and only passes the serial-fallback string through. The reply-path offset arithmetic (`63 + ssid_len : 69 + ssid_len`) and the invalid-MAC rejection are unverified. Consider adding a focused `_extract_mac` unit test covering all three branches.

wuwentao (commented):
### Review: MAC address reporting (event: COMMENT)

Thanks for adding MAC reporting. The parameter threading through `device_selector`, `MideaDevice.__init__`, and the 37 per-device `__init__.py` files is consistent and backward-compatible (the new `mac` kwarg defaults to `None`, the new dict key is additive), and `mac` is defined on every return path.

Two non-blocking findings, both in `midealocal/discover.py` (see inline comments). Lint/format/type checks were intentionally left to CI and are not covered here.

- **Medium** - `mac` value is inconsistent (12-char lowercase MAC vs 16-char serial fragment) depending on source.
- **Low** - `_extract_mac` lacks direct unit tests for the reply and invalid-MAC paths.

wuwentao:
> I don't have the vscode+devcontainers right now, but the CI is now passing on my fork so I guess it's ready for another review

thanks for your update.
seems it's ready for reviw, so I just submit the review PR request to claude code and copilot cli, both post a different result as previous post. we can reveiw the result and check it with PR.

zeehio:
Thank you very much for the review.

I have addressed all the issues, including adding unit tests and fixing the `extract_mac()` function.

I'd appreciate another review round.

I will need to prepare another PR to expose the serial number in the device class (like I did with the mac in this PR), but I plan to submit that on a separate PR after merging this one.

Once these PR are done and midea-lan is released I'll move to the home assistant custom component to get both attributes and expose them.

chemelli74:
Hi,

We should add all those settings to a common class to derive each device specific one from in order to avoid repeating and repeating same code and have consistency granted.

zeehio:
As far as I remember they are already in a common class, it just happens we are passing lots of arguments around to each of the >20 device types. For each device type, instead of receiving lots of arguments from the base device class and then calling `super().__init__()` with all those arguments, we could pass the kwargs with everything.

I don't have a Claude Code subscription but it seems to me that it should be easy for Claude code to change that.

My proposal to you would be to (1) get this PR reviewed and merged (2) ask Claude to evaluate the argument refactor to the device classes so we do not need to explicitly pass all the arguments every time and (3) add the serial_number to the base device class.

In case you think it's a good roadmap, and since I don't have an AI code subscription, I would appreciate if you could handle (2) and then I handle (3) again.

If you have other plans I'm fully open. This is just a suggestion and I may have missed something.

zeehio:
If you want to review and merge #491 first, I can rebase this PR afterwards. It removes a lot of boilerplate code passing init arguments to the parent class

midealocal/device.py
```diff
@@ -634,6 +641,10 @@ def set_ip_address(self, ip_address: str) -> None:
             self._ip_address = ip_address
             self.close_socket()
 
+    def set_mac(self, mac: str | None) -> None:
```
wuwentao, line 658: **Observation:** `set_mac` is never called in the codebase (MAC is set exclusively via the constructor kwarg). You added `test_set_mac`, suggesting it's intentional public API surface for external consumers. If so, this is fine — if not, consider removing it as dead code. Also note: the body `self._mac = mac or None` normalizes `""` to `None`, which differs from `__init__` (line 113: `self._mac = mac` stores `""` verbatim). Consider matching the constructor's behavior: `self._mac = mac`.

midealocal/discover.py
```diff
@@ -226,6 +252,8 @@ def _parse_discover_response(
             model = sn[3:11]
         else:
```
wuwentao, line 259: **Feature parity:** The protocol-1 (XML) path hardcodes `mac = None` (line 253), while protocol-2/3 paths extract MAC from the SN if available. For consistency, consider also attempting SN extraction here when `len(sn) >= 28`:

```python
        # No mac available in XML response; try serial number fallback
        if len(sn) >= 28:
            mac = sn[16:28]
            if mac and not re.fullmatch(r"[0-9A-Fa-f]{12}", mac):
                _LOGGER.warning("Invalid MAC address %s", mac)
                mac = None
        else:
            mac = None
```
(Or factor out the SN-extraction + validation logic into a helper to avoid duplication.)

wuwentao (commented):
Great improvements in this revision! You've addressed all the critical findings from my earlier review:

- ✅ Fixed the 16-char MAC bug (now `sn[16:28]` for 12 chars)
- ✅ Regex validator now checks length (`{12}`)
- ✅ Docstring typo fixed ("MAC" not "MAX")
- ✅ E501 line-length fixed (`# noqa: E501`)
- ✅ Magic numbers improved (named `mac_start`/`mac_ends`)
- ✅ Redundant `bool()` removed
- ✅ Dead `sn is not None` check removed
- ✅ Test coverage added (`test_extract_mac` in cli_test.py)

Two minor observations inline (not blockers). I also have one small correctness fix prepared as a commit (the `len(sn) >= 28` change for multibyte UTF-8 edge case) — see separate message for details.

midealocal/device.py
zeehio: I added the test because I may need this method from the home assistant integration afterwards .

Normalisation "" to None also happens in the base class for consistency 

https://github.com/zeehio/midea-local/blob/257bb52c197c1bb95a200be77386b905ceb6bc2a/midealocal/device.py#L113

midealocal/discover.py
zeehio: I don't have any device using that protocol. I found an example of that protocol on a internet forum, but it included a 22-character serial number. No MAC address was included in that serial number. Therefore it is not always possible to obtain the MAC address from the device.

If we do not have the Mac address then home assistant won't be able to use it, but that's ok.

zeehio: And without a real example of a longer serial number I don't dare to guess where the Mac address may be

zeehio:
> I also have one small correctness fix prepared as a commit (the `len(sn) >= 28` change for multibyte UTF-8 edge case) — see separate message for details.
> 

Does this change make any sense at all? Serial numbers are as far as I know either 32 or 22 ASCII characters. Checking for a 28 length serial number is confusing and talking about multibyte utf-8 sequences on an ascii serial number is also meaningless.

What do you think?

wuwentao:
> > I also have one small correctness fix prepared as a commit (the `len(sn) >= 28` change for multibyte UTF-8 edge case) — see separate message for details.
> 
> Does this change make any sense at all? Serial numbers are as far as I know either 32 or 22 ASCII characters. Checking for a 28 length serial number is confusing and talking about multibyte utf-8 sequences on an ascii serial number is also meaningless.
> 
> What do you think?

agreen with it, AI just our tool and help us, we still need to review and check it. we can ignore it, thanks for your help and support.

zeehio:
> > > I also have one small correctness fix prepared as a commit (the `len(sn) >= 28` change for multibyte UTF-8 edge case) — see separate message for details.
> > 
> > 
> > Does this change make any sense at all? Serial numbers are as far as I know either 32 or 22 ASCII characters. Checking for a 28 length serial number is confusing and talking about multibyte utf-8 sequences on an ascii serial number is also meaningless.
> > What do you think?
> 
> agreen with it, AI just our tool and help us, we still need to review and check it. we can ignore it, thanks for your help and support.

Great! I can clean this up once the refactor is merged. I will add comments and checks  to prevent the AI (and any future reader) making wrong assumptions.

midealocal/device.py
zeehio: __init__ now uses set_mac as well so handling empty macs as None happens in one and only one place.  I believe this is resolved

midealocal/discover.py
zeehio: We have test coverage now. Resolved

midealocal/discover.py
zeehio: We can extract the Mac address if the serial number is of type1. This is the best that we can do. Resolved

zeehio:
Rebased and ready to review

wuwentao (approved):
## Re-review — all feedback addressed ✅

Re-reviewed at head `d74ee9a` (rebased onto the merged #491 kwargs refactor). Every earlier review comment is resolved, and I verified the changed code with independent checks. CI is green across Python 3.12/3.13/3.14 × macOS/Ubuntu/Windows.

### Earlier comments — all resolved
| Comment | Resolution (verified) |
|---------|----------------------|
| `mac` format/length inconsistent (16-char SN slice vs 12-char reply) | **Fixed** — SN fallback is now `sn[16:28]` (12 chars), and the guard is `re.fullmatch(r"[0-9A-Fa-f]{12}", mac)` which enforces *both* hex and length. Both extraction paths yield exactly 12 chars. |
| No test coverage for `_extract_mac` | **Fixed** — `test_extract_mac` covers reply-offset extraction, SN fallback, and the invalid→None guard. |
| `set_mac` unused / `""`-vs-`None` inconsistency | **Resolved** — `__init__` now calls `set_mac(kwargs.get("mac"))`, so `"" → None` normalization lives in exactly one place; `set_mac` is intentional public API for the HA integration. |
| Protocol-1 (XML) MAC parity | **Resolved** — protocol-1 now calls `_extract_mac(b"", 0, sn)` too, extracting from a type-1 serial (and correctly returning `None` for the type-2 serials you found in the wild). |
| #491 integration | **Done cleanly** — `mac: NotRequired[str | None]` added to `MideaDeviceInitKwargs`; the 40 per-subclass edits are gone. Exactly the right shape. |

### Verification results (changed code)
- **`_extract_mac`** — both MAC sources are exactly 12 chars; the `{12}` regex length-anchors and hex-validates (nulls non-hex serial slices, always passes lowercase `.hex()`); no off-by-one in the `reply[mac_start:mac_ends]` slice; the protocol-1 `_extract_mac(b"", 0, sn)` call is safe and cannot crash (both call sites pass a decoded `str`).
- **`set_mac`/kwargs** — `mac` is optional end-to-end (`NotRequired` + `kwargs.get("mac")`, no `KeyError`); normalization unified; `set_mac` is the last statement in `__init__`, so no read-before-assign.
- **`utf-8` → `ascii` decode** — a good change: it makes `len(sn) == 32` reliable (1 byte = 1 char), which the earlier multibyte concern depended on. Not a crash regression — `utf-8` already raised uncaught `UnicodeDecodeError` on invalid bytes.

### One pre-existing note (not blocking, not this PR)
`discover()`'s receive loop catches only `TimeoutError`/`OSError`, so a `UnicodeDecodeError` from a single malformed reply would abort the whole scan. This **predates** #488 (the prior `utf-8` decode had the same exposure). Worth a small follow-up (`except UnicodeDecodeError: continue`), but out of scope here.

### Verdict: **approve** — ready to merge
All review feedback is addressed, the changed code is correct, and CI is green. Nice work, @zeehio. 🎉

_Review processed by Claude Code (Opus 4.8)._

PR #491, opened 2026-07-02
refactor: use kwargs in device subclasses

zeehio:
If we add an argument to the base device class we
do not want to edit the +30 device subclasses.

This is related to #488, to remove boilerplate code and reduce the size of that PR

### DIFF
--- a/midealocal/device.py
+++ b/midealocal/device.py
@@ -6,7 +6,7 @@ import threading
 import time
 from collections.abc import Callable
 from enum import IntEnum, StrEnum
-from typing import Any, ClassVar
+from typing import Any, ClassVar, TypedDict, Unpack
 
 from typing_extensions import deprecated
 
@@ -64,39 +64,53 @@ class MessageResult(IntEnum):
     ERROR = 99
 
 
+class MideaDeviceInitKwargs(TypedDict):
+    """Connection/identity kwargs forwarded by device subclasses to MideaDevice.
+
+    Every device subclass's ``__init__`` accepts these via ``**kwargs`` and
+    forwards them unchanged to ``MideaDevice.__init__``. Keeping them in one
+    place means adding a field here (e.g. ``mac``, ``serial_number``) is
+    enough for every subclass to accept and forward it, with no per-subclass
+    signature changes required.
+    """
+
+    name: str
+    device_id: int
+    ip_address: str
+    port: int
+    token: str
+    key: str
+    device_protocol: ProtocolVersion
+    model: str
+    subtype: int
+
+
 class MideaDevice(threading.Thread):
     """Midea device."""
 
     def __init__(
         self,
-        name: str,
-        device_id: int,
+        *,
         device_type: DeviceType,
-        ip_address: str,
-        port: int,
-        token: str,
-        key: str,
-        device_protocol: ProtocolVersion,
-        model: str,
-        subtype: int,
         attributes: dict,
+        **kwargs: Unpack[MideaDeviceInitKwargs],
     ) -> None:
         """Midea device initialization."""
         threading.Thread.__init__(self)
         self._attributes = attributes or {}
         self._socket: socket.socket | None = None
-        self._ip_address = ip_address
-        self._port = port
+        self._ip_address = kwargs["ip_address"]
+        self._port = kwargs["port"]
         self._security = LocalSecurity()
-        self._token = bytes.fromhex(token)
-        self._key = bytes.fromhex(key)
+        self._token = bytes.fromhex(kwargs["token"])
+        self._key = bytes.fromhex(kwargs["key"])
         self._buffer = b""
-        self._device_name = name
-        self._device_id = device_id
+        self._device_name = kwargs["name"]
+        self._device_id = kwargs["device_id"]
         self._device_type = device_type
-        self._device_protocol_version = device_protocol
-        self._model = model
-        self._subtype = subtype
+        self._device_protocol_version = kwargs["device_protocol"]
+        self._model = kwargs["model"]
+        self._subtype = kwargs["subtype"]
         self._message_protocol_version: int = 0
         self._updates: list[Callable[[dict[str, Any]], None]] = []
         self._unsupported_protocol: list[str] = []
--- a/midealocal/devices/a1/__init__.py
+++ b/midealocal/devices/a1/__init__.py
@@ -3,10 +3,10 @@
 import json
 import logging
 from enum import StrEnum
-from typing import Any, ClassVar
+from typing import Any, ClassVar, Unpack
 
-from midealocal.const import DeviceType, ProtocolVersion
-from midealocal.device import MideaDevice
+from midealocal.const import DeviceType
+from midealocal.device import MideaDevice, MideaDeviceInitKwargs
 
 from .message import MessageA1Response, MessageQuery, MessageSet
 
@@ -55,29 +55,14 @@ class MideaA1Device(MideaDevice):
 
     def __init__(
         self,
-        name: str,
-        device_id: int,
-        ip_address: str,
-        port: int,
-        token: str,
-        key: str,
-        device_protocol: ProtocolVersion,
-        model: str,
-        subtype: int,
+        *,
         customize: str,
+        **kwargs: Unpack[MideaDeviceInitKwargs],
     ) -> None:
         """Initialize Midea A1 device."""
         super().__init__(
-            name=name,
-            device_id=device_id,
             device_type=DeviceType.A1,
-            ip_address=ip_address,
-            port=port,
-            token=token,
-            key=key,
-            device_protocol=device_protocol,
-            model=model,
-            subtype=subtype,
+            **kwargs,
             attributes={
                 DeviceAttributes.power: False,
                 DeviceAttributes.prompt_tone: True,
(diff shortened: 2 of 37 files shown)


wuwentao (commented):
## Review: `refactor: use kwargs in device subclasses`

### Overview
Replaces the repeated 9-parameter connection/identity signature in all 36 device subclass `__init__` methods with `def __init__(self, *, customize: str, **kwargs: Unpack[MideaDeviceInitKwargs])`, forwarding `**kwargs` to `MideaDevice.__init__`. A new `MideaDeviceInitKwargs` TypedDict centralizes the forwarded fields. Net −519 lines. This is the *altitude* fix for the boilerplate #488 exposed (adding `mac` there required editing all 36 subclasses).

### Code quality — strong ✅
- Clean, uniform transformation across all 36 subclasses. Verified: no subclass referenced a connection param in its body, none had extra params, all follow the identical pattern.
- Keyword-only (`*`) is the right call — prevents positional-arg mistakes and makes the `customize` / `**kwargs` split unambiguous.
- Correctly drops the now-unused `ProtocolVersion` import from each subclass (verified: 0 remaining references in sampled files).
- `# noqa: ARG002` on `customize` is preserved for subclasses that don't use it — no lint regression.
- `device_type` and `attributes` stay explicit per-subclass (correct — device-specific, not part of the forwarded set), so no keyword collision with `**kwargs`.

### Relationship to #488 (you asked about this)
Both PRs touch all 36 subclass files → **guaranteed merge conflicts**. Recommended: **merge #491 first**, then rebase #488 to its minimal form — after that, #488 only needs to add `mac` to `MideaDeviceInitKwargs` + the `mac` param to `MideaDevice.__init__`, and the 36 per-subclass `mac=mac` edits disappear. See the inline comment on the TypedDict for the exact integration (note: `mac` must be `NotRequired` since it's optional).

### Risks / to verify
- **mypy (PEP 692) is the key gate.** Config is strict and mypy runs in pre-commit. The `Unpack[MideaDeviceInitKwargs]` forwarding should type-check on modern mypy, but confirm the CI mypy job is green — it's the load-bearing assumption of the whole refactor.
- **Keyword-only is a public API change:** external code constructing a device *positionally* would break. In-repo, `device_selector` and all tests use keywords, so nothing breaks here; the HA integration also goes through `device_selector`, so it should be fine.

### Test coverage
Existing device tests construct with all-keyword args (verified in the a1 test), so they're compatible — no changes needed. A single assertion that `device_selector(...)` still builds each device type would guard the forwarding chain, but it's optional.

### Security
No concerns — pure signature refactor, no I/O or data-handling changes.

---
Overall: clean, well-motivated refactor that meaningfully reduces maintenance surface. Main action is confirming the mypy CI gate and coordinating merge order with #488.

midealocal/device.py
```diff
@@ -64,6 +64,27 @@ class MessageResult(IntEnum):
...
+
+    name: str
+    device_id: int
+    ip_address: str
+    port: int
+    token: str
+    key: str
+    device_protocol: ProtocolVersion
+    model: str
+    subtype: int
```
wuwentao, line 85: **Nice centralization** — this is the right fix for the per-subclass boilerplate #488 ran into.

**#488 integration:** when #488 rebases onto this, `mac` should be added *here* instead of to 36 subclasses. Since #488's `mac` is optional (`mac: str | None = None`), it must be `NotRequired`, otherwise it becomes a required kwarg and every `device_selector` call and device test would fail under this repo's strict mypy config:

```python
from typing import NotRequired  # stdlib, 3.11+

class MideaDeviceInitKwargs(TypedDict):
    name: str
    device_id: int
    ip_address: str
    port: int
    token: str
    key: str
    device_protocol: ProtocolVersion
    model: str
    subtype: int
    mac: NotRequired[str | None]  # optional; added by #488
```
…and `MideaDevice.__init__` gains `mac: str | None = None`. That's the entire subclass-facing change #488 needs after this lands — the 36 `mac=mac` edits disappear.

**Minor (sync):** these 9 fields duplicate `MideaDevice.__init__`'s signature (lines 93-102), so the two must be kept in sync by hand. mypy will flag drift at the `super().__init__(**kwargs)` call sites, so it's low-risk — but worth a comment linking them, or consider having `MideaDevice.__init__` itself take `**kwargs: Unpack[MideaDeviceInitKwargs]` for a single source of truth.

midealocal/devices/a1/__init__.py
```diff
@@ -55,29 +55,14 @@ class MideaA1Device(MideaDevice):
...
-        ip_address: str,
-        port: int,
-        token: str,
-        key: str,
-        device_protocol: ProtocolVersion,
-        model: str,
-        subtype: int,
+        *,
         customize: str,
+        **kwargs: Unpack[MideaDeviceInitKwargs],
```
wuwentao, line 60: **Pattern looks correct** and is applied uniformly across all 36 subclasses (I checked: no subclass referenced a connection param like `model`/`device_id` in its body, and none carried extra params, so nothing is lost by folding them into `**kwargs`). Keeping `customize` explicit — it's device-specific and consumed per-subclass — while forwarding the connection/identity params is the right split, and keyword-only (`*`) prevents positional-arg mistakes.

**Key CI gate:** `super().__init__(device_type=..., **kwargs, attributes=...)` must type-check under the strict mypy settings (`disallow_untyped_calls`, `warn_unreachable`, `python_version = 3.14`). PEP 692 `Unpack[TypedDict]` forwarding is supported on modern mypy, but since it's the load-bearing assumption for all 36 files, a green mypy run is the gate to confirm before merge.

midealocal/device.py
zeehio: I can make MideaDevice.__init__() to take kwargs, but some of the argument order may change. If I follow that road I would make the constructor take only keyword arguments, and error if used on positional arguments. This may break some downstream users, but it is safer considering the amount of arguments the constructor already has.

Would you like me to do that?

wuwentao: @zeehio thank you very much for your quickly response, and I will have some more delay, as i just run these  claude jobs in background and not check it.
anyway, for this review comments, we can update `MideaDevice.__init__ itself take **kwargs: Unpack[MideaDeviceInitKwargs] for a single source`.
could you help to update it again, thanks.

midealocal/devices/a1/__init__.py
zeehio: This thread doesn't require any further action. If the CI passes then we are ready

midealocal/device.py
zeehio: Fixed

wuwentao:
## Re-review — `refactor: use kwargs in device subclasses` ✅

Re-reviewed at head `eeea1db` after the updates. **The review comments were addressed well — this looks ready to merge.**

### What changed since my last review
My earlier inline note flagged that the `MideaDeviceInitKwargs` TypedDict *duplicated* `MideaDevice.__init__`'s parameter list, needing manual sync. The author resolved this **better than I suggested**: `MideaDevice.__init__` itself now consumes `**kwargs: Unpack[MideaDeviceInitKwargs]` and reads fields via `kwargs[...]`, so the TypedDict is now the **single source of truth** — the signature is no longer duplicated anywhere.

### Verification (all ✅)
- **Uniformity:** all 36 subclasses converted to `def __init__(self, *, customize: str, **kwargs: Unpack[MideaDeviceInitKwargs])` + `super().__init__(device_type=..., **kwargs, attributes={...})`. Verified none lost body logic, and none carry constructor params beyond `customize`.
- **No keyword collision:** `MideaDevice.__init__` is `(*, device_type, attributes, **kwargs)`; subclasses pass `device_type=` and `attributes=` explicitly while `**kwargs` carries only the 9 connection fields — no overlap.
- **Init ordering is correct:** `super().__init__(**kwargs)` runs first (setting `self._subtype` etc.), *then* subclass bodies use it — e.g. `ac` calls `self.set_customize(customize)` and later reads `self._subtype`, all after super. No use-before-set.
- **`customize` handling untouched:** subclasses that use it keep the plain param (e.g. `ac`), those that don't keep `# noqa: ARG002` — all pre-existing context, no regression.
- **Callers satisfy the keyword-only signature:** `device_selector` passes all fields by keyword; the direct `MideaDevice(...)` construction in `tests/device_test.py` does too.
- **Imports cleaned:** all 36 subclasses correctly dropped the now-unused `ProtocolVersion` import (→ `import DeviceType` only); no F401.
- **CI:** green across Python 3.12 / 3.13 / 3.14 × macOS / Ubuntu / Windows — including the **strict mypy** gate (`disallow_untyped_defs`, `warn_unreachable`, `python_version = 3.14`). This was the key risk I flagged (PEP 692 `Unpack[TypedDict]` support), and it passes.

### Net effect
−526 lines of boilerplate, and future base-class connection fields now require editing exactly **one** place (the TypedDict + base `__init__`) instead of all 37 files — which is precisely the maintenance win #488 motivated.

### Minor note (non-blocking, no change requested)
With `kwargs["port"]` etc., an *untyped/dynamic* caller that omits a field would get a runtime `KeyError` rather than the old `TypeError: missing required argument`. In practice this is a non-issue — mypy guards every real caller (CI proves it) and this is an internal library constructor — so I'd leave it as-is. Flagging only for awareness.

### Recommendation
**Approve / ready to merge.** Clean, well-motivated refactor; review feedback incorporated; CI green; no correctness or convention issues found.

PR #493, opened 2026-07-03
feat: report serial number

zeehio:
Similar but easier to #488 here we want the device to report the serial number, so we can expose it as well in home assistant


midealocal/device.py
```diff
@@ -668,7 +668,9 @@ def set_mac(self, mac: str | None) -> None:
 
     def set_serial_number(self, sn: str | None) -> None:
```
chemelli74: Why not make it a property ? Looks more correct to me

chemelli74 (dismissed):

midealocal/device.py
zeehio: I have kept the style of the class as it is currently. There are getter properties (e.g. self.serial_number) but not setter properties. Check `set_ip_address`, `set_available`, `set_mac`...

Would you rather have a mix of  "setter properties" and "setter methods"?

chemelli74: I would add serial number as property for this PR and then in a separate PR convert current methods to property and have a clean implementation

zeehio: Done

zeehio:
@chemelli74 I defined the setter as a property as you requested.

I got some false positives with mypy and Claude worked around them. I kept the commit separate in case you want to take a look.

If there's anything else you need feel free to tell me. I'd be happy to see this pr merged :-)

chemelli74:
Honestly I'm don't see the need for a setter but I may miss something from the phone.

Why not set self._serial and then have the property just return it ?

wuwentao:
> **⚠️ Update / correction — superseded.** I posted the summary below before I had the full review thread in view, and I have since **dismissed my approval**. The branch has advanced: @zeehio converted the setter into a `@property` setter (addressing the original "make it a property" request), and the remaining design question — property setter vs. no setter (per @chemelli74's latest note) — is still an open discussion between author and reviewer. Please **disregard the "approve / ready to merge" conclusion** below. CI is green on the current head, but the `CHANGES_REQUESTED` review is intentionally left in place until that discussion resolves.

<details><summary>Original summary (kept for history — superseded)</summary>

## Review summary — `feat: report serial number`

Reviewed following the `review` skill, with each finding independently verified by parallel sub-agents. The PR is well-scoped and cleanly mirrors #488 (`feat: report MAC address`). Full CI matrix (3.12–3.14 × macOS/Ubuntu/Windows) is green.

### Findings

- **[VERIFIED SAFE] `cli.py:124` `serial_number=device["sn"]`** — the `"sn"` key is unconditionally present in the single device-dict literal returned by `_parse_discover_response` (both the V3 and V1/XML branches set it), so this cannot `KeyError`. Same guarantee as the adjacent `device["mac"]`.

- **[FIXED] `__init__` bypassed the setter** — `__init__` initialized `mac` via `self.set_mac(...)` but assigned `self._serial_number` directly. Routed through the setter to match `mac`.

- **[FIXED] No empty/NUL normalization** — `set_mac` normalizes `"" → None`, but the serial stored the raw value. Discovery can yield an empty `apc_sn` (XML branch) or a NUL-padded fixed-width V3 serial (`reply[8:40]`), so `serial_number` could surface `""` or `"\x00…"` where `mac` would report `None`. Now normalized to `None`.

- **[ADDRESSED] Test coverage** — extended the serial-number test to cover empty, NUL-padded, `None`, and NUL-trailing serials.

**Overall assessment: approve.** (Superseded — see the correction above.)

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code) (Opus 4.8)</sub>

</details>

wuwentao (dismissed):
Approving — ready to merge. Serial-number reporting is correct and now symmetric with the `mac` attribute (empty/NUL serials normalize to `None`). Full CI matrix (3.12–3.14 × 3 OSes) green on the latest commit. See review summary above for details.

zeehio:
No setter method. Done. There was a lot of back and forth so I squashed the whole thing to a single simple commit. @chemelli74 I hope this is now what you wanted

wuwentao:
@chemelli74 
as you have add changes request for most of the opening PRs in current repo,  could you help to review and confirm the latest result?

in addition, per my understand, as a repo owner/maintainer, for PR review, if PR owner has no any response for a long time and the opening PR only have some minor/small issue, most of the feature works well with code review, and we want to support new features and quickly fix bug/issue,  we can help them to quickly fix it or accept some minior issue in exist PRs or try to fix it with a new PR, then we can quickly merge more PRs, thanks
> not for this PR, as @zeehio always post commit/response quickly. 


thanks

wuwentao (approved):

wuwentao:
hi @chemelli74 
if there is no any response, I will try to ignore exist review comments feedback and merge these PRs.
as I need to publish a new release for midea-local and also publish a new release with midea_ac_lan.

anyway, if there is something missed during PR review, we still can submit new PR to fix it in future.

chemelli74:
@wuwentao, we work on the code on our free time. Is not a job.

I waited in the past so many days to ear from you. I think you can wait a couple from me isn't it ? ;-)

midealocal/device.py
```diff
@@ -125,6 +126,10 @@ def __init__(
         self._previous_heartbeat = 0.0
         self.name = self._device_name
         self.set_mac(kwargs.get("mac"))
+        # Discovery may report a fixed-width serial padded with NUL bytes or an
+        # empty ``apc_sn`` attribute; normalize these to None (mirrors set_mac).
+        sn = kwargs.get("serial_number")
+        self._serial_number = (sn.strip("\x00").strip() or None) if sn else None
```
chemelli74, line 134:
```suggestion
        sn = kwargs.get("serial_number", "\x00")
        self._serial_number = sn.strip("\x00").strip() or None
```

zeehio: This suggestion has two issues:

- When `kwargs["serial_number"]  = None`, `sn`  becomes None and the strip method raises AttributeError. I believe this reason is sufficient to reject the suggestion as is

- When kwargs does not contain a "serial_number", sn becomes "\x00" and afterwards that "\x00" is stripped. It would be simpler to use kwargs.get("serial_number", "") instead. 

You fail to say what is it that you don't like from the implementation I provided -or in other words why you suggest the change- so I have to guess what is it that you don't like and try to fix it.

Maybe you would like better something like:

```python

sn = kwargs.get("serial_number") or ""
self._serial_number = sn.strip("\x00").strip() or None

```
This would be equivalent to my implementation and maybe you find it more readable.

I like the detailed code review but I would appreciate more clarity in your feedback.

zeehio:
@chemelli74 please take this comment as constructive feedback. I think it is positive for the project for you to review these pull requests.

I appreciate your review because from it we determined we did not need a `set_serial_number` method, reducing the maintenance surface of the package. That was a good outcome of your review. I'm thankful. It makes the package better.

However I would also appreciate more clarity in your feedback. I believe you could be more clear in the things you don't like or you would rather have implemented differently. I'll give you two examples from this pull request hoping that you can understand my point of view, even if you may disagree with it.

See for instance the first review:

- My [first submission](https://github.com/midea-lan/midea-local/commit/f1f6c65331239fbba539b0556341c0cbade0068e) made serial_number a read only property with a set_serial_method that wasn't part of the property (it wasn't decorated with `@serial_number.setter`). I did that to be consistent with all the other properties in this class.
- You then highlighted the `set_serial_number` method and suggested to make `serial_number` a property.
- I *interpreted* then that you wanted me to decorate that method as a setter, and I told you so.
- You validated my interpretation and I implemented that change.
- Then you stated the setter method was not needed at all.
- Finally I removed the setter method. I forced pushed so we do not end up having commits going back and forth with a setter method that is eventually removed.

We reached a good final state, but (1)  I had to guess what you meant and (2) you changed your opinion afterwards. Of course you are entitled to (2), but (1) in my opinion shows some lack of clarity on your end.

And another example:

This final code suggestion has no clear statement on what are you aiming to achieve with the change. It also happens that your code suggestion has a flaw. So I have to assume there is something you don't like in my code (because you suggested a change) and I am forced to guess what is it that you don't like in my submission. Your feedback makes me assume and guess. Again, clarity on what you like and what you don't would help here.

I have spent a significant time/effort writing this message because I really appreciate your feedback, even if -in my opinion- its clarity could be improved. I hope you can accept it.

I'll be happy to amend my submission once I understand what is it that you don't like about it.

chemelli74:
> @chemelli74 please take this comment as constructive feedback. I think it is positive for the project for you to review these pull requests.

You can join me on Discord so we can discuss more in depth ;-)


EDIT:

> You validated my interpretation and I implemented that change.

I re-checked and cannot find a comment where I validated the settler.
Can you point me to it please ?

midealocal/device.py
chemelli74: Your code does 2 lookups for `None`, I don't like it because is not readable.

If you first default to `\00` when serial_number is missing, then you can just strip and that's it.
No more if needed or defaults to care about.

wuwentao:
> @wuwentao, we work on the code on our free time. Is not a job.
> 
> I waited in the past so many days to ear from you. I think you can wait a couple from me isn't it ? ;-)

Thanks for your reply; we have also been waiting for your feedback.
There are currently a large number of pending issues and PRs. Since I wasn't sure when you might respond, I simply wanted to point out that we cannot keep waiting indefinitely if there is no response after several weeks.
My main concern is that the PR owner might not have the time later to provide further feedback or update the status, leaving things in a perpetually pending state.

I fully understand that you want to ensure high code quality and robust functionality. However, we should be willing to accept minor imperfections and encourage others to contribute code and features. After all, everyone has different habits and coding styles; during the review process, we can simply state our expectations, with the goal of reaching a consensus with the PR owner as quickly as possible.

Personally, I don't think we should require every person or every PR to adhere strictly to our own ideas or coding styles. As long as the code's functionality and logic meet actual usage requirements, that should suffice. Bugs are inevitable, after all—no one can solve every single problem.

midealocal/device.py
zeehio: > Your code does 2 lookups for None, I don't like it because is not readable.

Thank you for this answer.

> If you first default to \00 when serial_number is missing, then you can just strip and that's it.

Or just default to an empty string? Why do you think it is better to default to \00 instead and then stripping it?

> No more if needed or defaults to care about.

I'm defensively considering two scenarios: (1) serial_number is missing from kwargs and (2) serial_number is not missing from kwargs but it is set to None.

Your suggestion fails to consider  (2) as I stated in my first point in the previous comment. Do you acknowledge that limitation in your suggested code?

Therefore do you and agree with me that your suggestion needs to be changed?

Do you find the code suggestion I wrote in the previous comment readable enough?

Do you agree that it covers both (1) and (2)?

If you  don't like the code I suggested, can you suggest an alternative to your liking?

It's a annoying on my end to have to resort to typing these fine grained questions. I do this because your reply fails to share your position with respect to my points while I am doing my best to agree or disagree with each of your points in your comments. I won't do this again. Either you improve your communication being more assertive or I will ask someone else to step in.

zeehio:
> > @chemelli74 please take this comment as constructive feedback. I think it is positive for the project for you to review these pull requests.
> 
> You can join me on Discord so we can discuss more in depth ;-)

I don't really want to discuss this more, much less in another platform. I'm here to make my AC report its serial number and I'm trying to workaround through your communication skills hoping to help you in the process. So far I see zero progress or self-reflection on your end, so I have zero incentive to help you further in this aspect.

> 
> EDIT:
> 
> > You validated my interpretation and I implemented that change.
> 
> I re-checked and cannot find a comment where I validated the settler. Can you point me to it please ?

Sure.

First here https://github.com/midea-lan/midea-local/pull/493#discussion_r3527706078 you say you want serial_number to be a property. Since the getter serial_number already was a property and you were commenting on the setter method, I replied focusing on setter property methods.  Then you said

> I would add serial number as property for this PR and then in a separate PR convert current methods to property and have a clean implementation

There are two possible interpretations of this comment:

(a) You are referring to properties just with a getter. 
(b) You are referring to properties with both getter and setter.

Interpretation (a) doesn't make sense to me, because `serial_number` already had a getter property, as well as `mac` and others. So what you were asking was already there, at least for the `serial_number` under review but also for other properties.

Interpretation (b) makes more sense to me. You have seen the getter methods and you want more pythonic properties that include the setter as well; not as a conventional method but decorated as a part of the property. You are also replying to my comment, where I was focusing on the setter component of the property, and while you didn't explicitly agree or disagree to my comment you mentioned you wanted changes in the code in order to have properties. Since the getter methods of the properties are already implemented as properties you must be referring to the setter methods.

The alternative was that you didn't see in the code the getter properties that already existed. I assumed you had read the code and seen them.

You may find I am assuming too much. I did. And I did it because your feedback is poorly constructed.

wuwentao:
@zeehio 
I am truly sorry. The requirement for the SN feature came from my own personal feedback when I initially reviewed your MAC address implementation. This has unfortunately caused you a lot of unnecessary extra work and issues, especially after you’ve already contributed several PRs.
However, since these are open-source projects on GitHub, everyone contributes in their spare time based on personal interest. We don’t have strict divisions of labor or formal responsibilities, so it’s natural that people have different habits and approaches. I hope this hasn’t caused you too much trouble or inconvenience.

wuwentao:
@chemelli74
We’ve already communicated multiple times, and I’ve always respected your personal habits and preferences. I certainly don’t want to create any unnecessary trouble or misunderstandings. While respecting your way of working, we also need to respect the PR owners and the value of their contributions — after all, this is just a small project.

**Current situation:**
Both `midea-local` and `midea_ac_lan` now have a backlog of 10–20 PRs. Many of them have already addressed your review comments, received secondary review and confirmation via Claude, and been tested and verified by the PR owners on their actual devices.

You’ve recently found some time to handle https://github.com/home-assistant/core/pull/151930 (I’ve also been receiving watchlist notifications), but you haven’t yet had a chance to review or reply to the PR comments on the midea-local side.
Code ultimately serves end users. No matter how elegant or beautiful the code is, if it doesn’t function properly on real devices or cannot be used by users, it may be less valuable than even “imperfect” code. On the other hand, code that starts out a bit rough can go through rapid iteration. With timely and efficient feedback, it can be continuously improved and optimized until it reaches a high standard. Nothing is perfect in the first attempt.

**Solution:**
I apologize in advance, but if the situation remains the same next week, I will have to proceed with merging and releasing the PRs while bypassing the pending review comments. If you still find anything unsuitable or uncomfortable, you’re always welcome to submit your own PR to make changes or to help the PR owners address the issues directly before merge. There are many ways to do this. Our goal is simply to solve problems and deliver value to users.

zeehio:
@wuwentao No problem at all. Don't feel sorry, it is not your fault. It's my first time dealing with someone who I believe is technically competent but is also a really bad team player (reasons above). I decided I am using this situation as an experiment "does it help if, in this situation, I provide detailed feedback of the behavioural issues I find?" (So far I have learnt that doing this is a waste of time, so next time I'll do something different)

I am not in a hurry (this is just for fun) and I am not taking things personally, so there is no need to feel sorry. I feel fine. 🙂

If anything I'm sorry for you. I'm here for a pull request. You are here maintaining the project so you are dealing with lots of PR like this one. My advice -if you want it- is to consider parting ways and forking. I don't see any intention of changing attitudes on their side so if things continue like this you may eventually lose motivation, I would lose it.

As a project maintainer you have a clear "provide feedback fast, iterate fast, treat well contributors so they keep contributing so the project is useful and keeps improving". The other approach "thorough review, long wait times, contributors can wait or leave, limited maitnainer availability" is also a valid one. However sharing the ownership of a project without a shared maintenance policy seems very challenging to me.

Anyway take care and we will see next week.

chemelli74:
Hi @zeehio,

First of all, I'm sorry that this discussion ended up being frustrating and that it caused you to spend time on work that ultimately wasn't moving in the right direction.

My intention with the review comment was to suggest a different approach rather than a modification of the implementation you had already started. In my mind, the expected path was to first convert `serial_number` into a read-only property (without a setter) and then, in a separate PR, apply the same pattern to the other attributes.

Looking back, I can see that my comment was not explicit enough. What seemed clear to me was obviously not clear from the review itself, and I understand how that could lead to a misunderstanding.

When review comments leave room for interpretation, I encourage contributors to ask for clarification either on GitHub or on our Discord channel. A quick discussion is often enough to align on the intended direction before spending significant time on an implementation. At the same time, reviewers should strive to make their expectations as clear as possible, and in this case I could have done better.

Every contribution is valuable, and I sincerely hope this experience has not discouraged you from contributing in the future.

That said, our goal is not only to make code work. We also need to keep the codebase easy to read, easy to maintain, free from unnecessary boilerplate and complexity, and reliable over time. Sometimes this means taking a more incremental approach or making preparatory changes before moving on to a larger refactor.

As an example, when I worked on bringing Midea LAN into Home Assistant Core, I first had to make changes to the underlying library before I could continue with the integration itself. Likewise, parts of the codebase such as the config flow required cleanup and deduplication before further improvements could be made. Those intermediate steps may not always seem directly related to the end goal, but they often help us arrive at a cleaner and more maintainable result.

PR #494, opened 2026-07-03
chore: migrate dev environment from devcontainer to uv

wuwentao:
## Summary

Migrates the developer environment from the **VS Code + Dev Container** (Docker) workflow to **[uv](https://docs.astral.sh/uv/) + a local `.venv`**. The devcontainer was hard to set up; uv is a single cross-platform install (Windows/Linux/macOS/WSL2), faster, and needs no Docker.

**Nothing about the published package changes** — `setup.py`, the three `requirements*.txt` files, `midealocal/version.py`, release-please, and PyPI trusted publishing are all untouched. This is purely a dev-env/CI/docs migration.

## What changed

**Removed**
- `.devcontainer/` (Dockerfile, devcontainer.json, lock)
- `scripts/sync_pyutils.sh` (dead; it only updated the devcontainer's pipx tools)

**Dev bootstrap**
- Rewrote `scripts/setup.sh` and added `scripts/setup.ps1` (Windows PowerShell): `uv venv --python 3.12` → `uv pip install -r requirements-all.txt pre-commit` → install the pre-commit hooks. Both assume uv is installed (docs cover per-OS install).

**CI (`.github/workflows/python-build.yml`)**
- `astral-sh/setup-uv@v6` + `uv python install ${{ matrix.python-version }}` replaces `actions/setup-python`; uv now manages the 3.12/3.13/3.14 interpreters.
- Deps via `uv pip install`; the `.venv` is put on `GITHUB_PATH` so `pre-commit`, `python -m build`, and `pytest` run in it (this repo has no `pyproject.toml`, so `uv run` can't auto-discover the env).
- Matrix (3.12/3.13/3.14 × Linux/macOS/Windows), pre-commit, build, pytest, codecov, and the release→PyPI publish job are all functionally unchanged.

**Editor + tooling**
- `.vscode`: point at `.venv`; switch tasks from `pip3`/`python3` to `uv`.
- Aligned the Python floor (was inconsistent): ruff `target-version = py312`, pylintrc `py-version = 3.12`, pyupgrade `--py312-plus`, mypy `python_version = 3.12`. `setup.py` was already `>=3.12`.
- `.gitignore`: ignore `node_modules/` + `package*.json` (the commitlint config the setup script installs).

**Docs**
- Rewrote `.github/CONTRIBUTING.md` and `.github/CONTRIBUTING.zh.md` around uv: per-OS uv install (macOS/Linux/WSL2/Windows), quick start, common commands, and China-mirror via `UV_DEFAULT_INDEX` / `UV_PYTHON_INSTALL_MIRROR` (replacing the old Docker `PIP_MIRROR_DOMAIN`/`APT_MIRROR_DOMAIN`).
- Added a **Development** section to `README.md`; created a Chinese **`README_hans.md`**; updated `AGENTS.md` build/test/lint commands to `uv run` and the recommended-env line.

## How contributors set up now

```bash
# install uv once (macOS/Linux/WSL2)
curl -LsSf https://astral.sh/uv/install.sh | sh
# then, in the repo:
./scripts/setup.sh          # Windows: scripts\setup.ps1
```

## Verification (locally, this branch)

- `scripts/setup.sh` bootstraps cleanly (venv + deps + pre-commit hooks + commitlint config).
- `python -m pytest ./tests/` → **243 passed**.
- `mypy midealocal` (3.12) clean · `ruff check .` (py312, pinned v0.6.9) clean · `pylint --rcfile=pylintrc midealocal` clean.
- `pre-commit run` (ruff / ruff-format / pyupgrade `--py312-plus`) pass with no file rewrites.
- `python -m build` produces `midea_local-6.10.0-py3-none-any.whl` + sdist (version read from `version.py` as before).

CI will confirm the full 3.12/3.13/3.14 × 3-OS matrix.


chemelli74 (dismissed):
I don't agree in removing the container.
Setup is one click and works for the majority of devs.

We can for sure introduce "uv", is on my todo list for a while ;-)

Check for example aioamazondevices  for the dev env setup

wuwentao:
> I don't agree in removing the container. Setup is one click and works for the majority of devs.
> 
> We can for sure introduce "uv", is on my todo list for a while ;-)
> 
> Check for example aioamazondevices for the dev env setup

it's only a python3 based lib, and no any external OS or CPU arch requirements, uv is enough for all the devs.
it can reduce build env, and CI also can sync with dev envs and CI envs.
as  in current mode, CI is a different dev mode with user devcontainer dev mode, we need to maintain two dev ENVs.

finally, we just remove the default devcontainer base image, if you still want to use devcontainer, there is no any limitation, you can use any linux-based OS  to do it, just run `setup.sh`, and the only requirement is uv runtime, we don't need to cost time to download docker base image or rebuild it when env changes.

chemelli74:
You can use uv without removing devcontainer.
They are not exclusive.

You start your system as you prefer and use uv in it.

I, and many others, build system via container that will use the same uv.

I don't see any limitation or issue with the above.

What's your concern ?

wuwentao:
> You can use uv without removing devcontainer. They are not exclusive.
> 
> You start your system as you prefer and use uv in it.
> 
> I, and many others, build system via container that will use the same uv.
> 
> I don't see any limitation or issue with the above.
> 
> What's your concern ?

for devcontainer, without it or with it, CI job and dev env can works well, this is not any concern, just why we need to keep it? anyway, as you still want to keep it in repo, so I have rebase the commit, please double check and confirm the result.

chemelli74:
>  as you still want to keep it in repo, so I have rebase the commit, 

Thx

> please double check and confirm the result.

I would like to have a script/setup.sh with all the logic so you can use it plain in your system or call it during devcontainer creation to get the env setup.
This way you have to maintain only that script.

Please take a look at https://github.com/chemelli74/aiocomelit for an example.

wuwentao:
> > as you still want to keep it in repo, so I have rebase the commit,
> 
> Thx
> 
> > please double check and confirm the result.
> 
> I would like to have a script/setup.sh with all the logic so you can use it plain in your system or call it during devcontainer creation to get the env setup. This way you have to maintain only that script.
> 
> Please take a look at https://github.com/chemelli74/aiocomelit for an example.


there is always a `scripts/setup.sh` exist in current projet and current PR.
and this PR only using `uv` to replace old `pip` command, if there is anything missed, could you help to add the detail? 
thanks

Issue #468, opened 2026-06-10
Midea CD water heater (RSJRAC07 / 170L HPWH): WiFi `set_temperature` faults the unit — `MessageSet` body is too short (8 bytes; firmware needs 25)


calebleemcd:
# Midea CD water heater (RSJRAC07 / 170L HPWH): WiFi `set_temperature` faults the unit — `MessageSet` body is too short (8 bytes; firmware needs 25)

## Affected device

Midea **CD** heat-pump water heater, type `0xCD` (205). Sold under many names — all the similar unit:

- **Trade Depot Midea 170L Hot Water Heat Pump** (NZ)
- **Chromagen / Ecospring 170L** (NZ/AU)
- Midea **RSJ-15/190RDN7** (= `RSJ-15/190RDN3-C`)
- Model code **`RSJRAC07`**, SN prefix `0000CD311RSJRAC07...`, subtype 0, protocol V3 (8370), new (raw-temp) Lua protocol.

Library: `midea-local` (tested 6.6.0), as used by `wuwentao/midea_ac_lan`. (Related: midea_ac_lan #515.)

## Symptom

Reads work fine over LAN. But **any WiFi `set_temperature` / `set_operation_mode` corrupts the unit**: the setpoint reads **0** and sticks, and the fault then poisons subsequent sets (even the official app's) **until a physical power-cycle**. Several owners have worked around this by disabling writes / using a separate RS485 Modbus tap.

## Root cause

`midealocal/devices/cd/message.py` `MessageSet._body` emits an **8-byte** body:

```
[0x01, power, mode, target, trValue, openPTC, ptcTemp, byte8]
```

The RSJRAC07 firmware expects a **25-byte** control body. The short body is structurally invalid and faults the device. (Both raw and `×2+30` target encodings of the short body fault — encoding was never the issue; **length/layout** is.)

### How the correct layout was found

The official app controls this unit via Midea **cloud** (the device holds a persistent TLS connection to `:28443`), so there is **no LAN set command to packet-capture**. Instead I downloaded + decrypted Midea's own Lua plugin for the device (`T_0000_CD_RSJRAC07_*.lua`) via the cloud API (`MideaCloud.download_lua`). Its `jsonToData` `controlType == 0x01` branch defines the real control body.

### The real 25-byte CD set body (msg_type `0x02`, body_type `0x01`)

[...]

PR #497, opened 2026-07-12
fix(cd): use 25-byte RSJRAC SET body with tsMax (#468)

RickyL1213:
## Summary

Fixes CD (heat-pump water heater) `controlType=0x01` SET frames so they no longer corrupt target/max temperature on **RSJRAC07** and related firmwares.

Closes / related: **#468**

### Problem

Stock `MessageSet` built a **22-byte** body ending at `body[21]`. RSJRAC07 firmware (Lua `T_0000_CD_RSJRAC07`) expects a **25-byte** control body. A short SET:

- zeros or clamps **target temperature** to 0 °C
- corrupts **max water temperature**
- can leave the unit stuck until power-cycle or restore via the official app

Symptom matches reports in #468 (and related integration issues such as wuwentao/midea_ac_lan#515).

### Root cause (from #468)

Official Lua `jsonToData` for `controlType == 0x01` defines:

| full idx | field | notes |
|---|---|---|
| 0 | body_type `0x01` | prepended by `MessageRequest` |
| 1 | constant `0x01` | |
| 2 | power | |
| 3 | mode | |
| 4 | target | raw °C (new protocol) |
| 5 | trValue | hysteresis [2, 6] |
| 6 | openPTC | Lua forces **0** on normal sets |
| 7 | ptcTemp | |
| 8 | flags | vacation / °F / mute |
| 9–20 | vacation/date | 0 for plain set |
| 21 | vacationTs | |
| 22 | timer_type | |
| 23 | **tsMax** | **must not be 0** (0 clamps setpoint) |
| 24 | dr_switch | |

Critical: **body length** and **non-zero `tsMax` at full[23]**. Putting max at full[21] (vacationTs) is wrong.

### Fix

- Extend `MessageSet._body` to **24-byte payload** (25 with body_type)
- Add `ts_max` from device `max_temperature`; fall back to `DEFAULT_TS_MAX` (65) if missing so we never emit 0
- Force `openPTC = 0` on normal sets (Lua)
- Clamp Tr into Lua range [2, 6]
- Plain power/mode/temperature sets leave `vacationTs = 0`; vacation paths still send vacation temperature/days
- Do not replay polluted SET-echo fields (`openPTC` / bad `trValue`) into the next frame

### Verification

Live HA → LAN (device **RSJRAC07**, protocol 3 / message protocol 8):

1. Set target to **63 °C** twice
2. SET echo shows target **63**, subsequent notify keeps **max 65**
3. Official app shows correct target/max (no 0 °C fault)

Unit tests cover body length, `tsMax` placement, Tr clamp, openPTC force-0, and vacation day encoding.

### Notes / risk

- Layout follows RSJRAC07 Lua; older CD subtypes with the short body may need gating if regressions appear
- After release, `midea_ac_lan` only needs a `midea-local` pin bump (no HA platform change)

## Test plan

- [x] Live set target 63 °C ×2 on RSJRAC07 via HA (`midea_ac_lan`)
- [x] Confirm app target/max stay non-zero
- [x] Unit tests for `MessageSet` 25-byte body + device `set_attribute` paths
- [ ] CI pytest on PR

### DIFF
--- a/midealocal/devices/cd/__init__.py
+++ b/midealocal/devices/cd/__init__.py
@@ -526,20 +526,29 @@ class MideaCDDevice(MideaDevice):
                 message.vacation_days = days
 
             # persist only safe fields; SET echoes often return openPTC=1 / bad Tr
-            fields = dict(message.fields)
-            for key in ("openPTC", "ptcTemp", "byte8"):
-                fields.pop(key, None)
-            tr = fields.get("trValue")
-            try:
-                tr_i = int(tr) if tr is not None else 0
-            except (TypeError, ValueError):
-                tr_i = 0
-            if tr_i < MessageSet.TR_VALUE_MIN or tr_i > MessageSet.TR_VALUE_MAX:
-                fields.pop("trValue", None)
+            self._fields = self._sanitize_set_fields(message.fields)
             # Avoid replaying SET-echo junk into the next control frame
-            message.fields = {k: v for k, v in fields.items() if k == "trValue"}
-            self._fields = fields
+            message.fields = dict(self._fields)
             self.build_send(message)
+
+    @staticmethod
+    def _sanitize_set_fields(fields: dict[Any, Any]) -> dict[Any, Any]:
+        """Strip SET-echo junk so it is not replayed into the next frame.
+
+        Drops openPTC/ptcTemp/byte8 (openPTC is forced 0 by MessageSet) and
+        removes an out-of-range trValue so only a valid Tr survives.
+        """
+        clean = dict(fields)
+        for key in ("openPTC", "ptcTemp", "byte8"):
+            clean.pop(key, None)
+        tr = clean.get("trValue")
+        try:
+            tr_i = int(tr) if tr is not None else 0
+        except (TypeError, ValueError):
+            tr_i = 0
+        if tr_i < MessageSet.TR_VALUE_MIN or tr_i > MessageSet.TR_VALUE_MAX:
+            clean.pop("trValue", None)
+        return {k: v for k, v in clean.items() if k == "trValue"}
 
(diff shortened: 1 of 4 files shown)


wuwentao:
## Review + fixes applied

I reviewed the 25-byte RSJRAC SET body change and it is functionally correct — the `full[23]` tsMax placement, non-zero fallback, Tr clamp, forced `openPTC=0`, and vacation-day encoding all match the issue #468 analysis. I ran the suite locally and pushed a follow-up commit (`60f6466`) to fix three issues that would have failed CI:

**1. Broken test (would error, not just fail)**
`test_set_target_temperature_body_length_and_ts_max` did `self.device.model = "RSJRAC07"`, but `model` is a read-only property → `AttributeError`. The line was also unnecessary: `set_customize('{"lua_protocol": "new"}')` already forces the new (raw °C) protocol regardless of model. Removed the assignment; the test now runs and passes.

**2. `C901` complexity (`set_attribute` was 21 > 18)**
The new field-sanitization block pushed `set_attribute` over the mccabe limit. Extracted it into a small `_sanitize_set_fields` static helper — same behaviour (strip `openPTC`/`ptcTemp`/`byte8`, drop out-of-range `trValue`, keep only `trValue`), now under the limit.

**3. Lint nits**
- `PLC0415`: moved the 5 in-function `MessageSet` imports to the module top-level.
- `D403`: capitalized the `test_ts_max_zero_falls_back` docstring.

### Local verification
- `pytest ./tests/` → **249 passed** (63 in `tests/devices/cd/`)
- `ruff check` + `ruff format --check` on the changed files → clean
- `mypy` (strict) on `devices/cd/__init__.py` → no issues

No behavioural change to the fix itself — these are test/lint corrections so the PR is green. LGTM after CI confirms. 🟢

PR #500, opened 2026-07-14
feat(ac): add power saving mode

### LINKED ISSUE #499: Add AC power saving mode support
## Device

- Device type: AC (`0xAC` / `172`)
- Model: `22012867`
- Subtype: `44204`
- Device protocol: `3`
- Message protocol: `8`

## Protocol evidence

The mode is exposed as Quiet Sleep in the MSmartHome app. Toggling it produces the following redacted local protocol observations:

| Operation | Frame | Observed state |
| --- | --- | --- |
| Enable | A0 notify | `power_saving=8`, `sleep_mode=False` |
| Enable | C0 query response | `power_saving=8`, `comfort_sleep_switch=64`, `sleep_mode=False` |
| Disable | C0 query response | `power_saving=0`, `comfort_sleep_switch=0`, `sleep_mode=False` |

Relevant redacted payloads:

```text
A0 enabled:  a05f40667f7f00000800000000000000000000f080480092dd
C0 enabled:  c0015b667f7f000008400070740f00610000000000000000580200005cff3d98
C0 disabled: c0015b667f7f00000000006c800f00580000000000000000580200005cff5880
```

The state is carried by `body[8] & 0x08`. The existing parser exposes the raw masked values `0` and `8`; it should be normalized to a boolean device attribute.

## Requested behavior

- Add a boolean AC `power_saving` attribute.
- Parse the `0x08` bit from A0 notifications and C0 query responses.
- Encode the same bit in X40 general-set messages.
- Preserve the current value during unrelated controls such as temperature or fan-speed changes.
- Include it in the existing mutually exclusive preset handling.
- Do not guess a write format for the unsupported AC subprotocol.

This behavior has also been validated on the physical device using a locally patched library: enable and disable both take effect, app-originated changes update the local state, and temperature/fan changes preserve the mode.

willylee007:
## Reason & Detail

Add protocol support for the AC `power_saving` state observed when toggling Quiet Sleep in MSmartHome.

- Add a boolean `DeviceAttributes.power_saving` attribute with a default value of `False`.
- Normalize A0 notifications and C0 query responses from the raw `0x00` / `0x08` mask to a boolean.
- Encode power saving as the `0x08` bit in X40 general-set messages without overwriting other bits in the same byte.
- Copy the current state into general-set messages so unrelated temperature, fan-speed, or swing controls do not disable the mode.
- Include power saving in the existing mutually exclusive AC preset handling.
- Reject writes for the unsupported AC subprotocol instead of guessing its protocol format.
- Add tests for parsing, encoding, bit composition, state preservation, mutual exclusion, and unsupported-subprotocol behavior.

Protocol evidence is documented in #499 for model `22012867`, subtype `44204`, device protocol `3`, and message protocol `8`.

Validation completed locally:

```text
pre-commit run --all-files: passed
pytest: 245 passed
python -m build: passed
physical device enable/disable: passed
app-originated state synchronization: passed
temperature and fan-speed state preservation: passed
```

## Real-device validation

The implementation was validated against a physical Midea AC through a Home Assistant Docker installation running on Synology NAS.

Device:

- Model: `22012867`
- Subtype: `44204`
- Device protocol: `3`
- Message protocol: `8`
- Integration: `midea_ac_lan`
- Protocol library: locally built `midea-local` with this change

Test results:

1. Enabling Quiet Sleep in the MSmartHome app generated an A0 notification with:
   ```text
   power_saving=8
   ```
   The Home Assistant entity updated to the Quiet Sleep preset.

2. A subsequent C0 query response reported:
   ```text
   power_saving=8
   comfort_sleep_switch=64
   sleep_mode=False
   ```

3. Disabling Quiet Sleep in the MSmartHome app generated a C0 response with:
   ```text
   power_saving=0
   comfort_sleep_switch=0
   sleep_mode=False
   ```
   The Home Assistant preset returned to `none`.

4. Enabling and disabling Quiet Sleep from Home Assistant successfully changed the physical device state.

5. While Quiet Sleep was enabled, changing the target temperature, fan speed, and swing settings preserved:
   ```text
   power_saving=8
   ```

6. The standard `sleep_mode` remained independent from `power_saving`; the two states were not confused during testing.

Only the normal AC A0/C0/X40 protocol path was validated. No write support is claimed for the unsupported AC subprotocol.

## Related issue

fix #499


midealocal/devices/ac/__init__.py
```diff
@@ -529,16 +532,24 @@ def set_attribute(self, attr: str, value: bool | int | str) -> None:
                 DeviceAttributes.self_clean,
             ]:
                 message = self.make_newprotocol_message_set(attr=attr, value=value)
+            elif attr == DeviceAttributes.power_saving and self._used_subprotocol:
+                _LOGGER.warning(
```
chemelli74:
```suggestion
                _LOGGER.debug(
```

tests/devices/ac/message_ac_test.py
```diff
@@ -464,6 +465,8 @@ def test_message_notify2_a0(self) -> None:
         assert hasattr(response, "swing_vertical")
         assert hasattr(response, "swing_horizontal")
         assert hasattr(response, "boost_mode")
+        assert hasattr(response, "power_saving")
+        assert response.power_saving is True
```
chemelli74, line 469: Note, we should assert the value for all attributes like it's done here for power_saving.

MAybe we should open a issue to remember us to do that (is there a tab "task" hopefully)

tests/devices/ac/message_ac_test.py
```diff
@@ -680,6 +683,8 @@ def test_message_query_c0(self) -> None:
...
```
chemelli74, line 687: Same as above.

And one more note: we can parametrize tests probably to avoid repeating same or similar code all over.

chemelli74 (dismissed):

midealocal/devices/ac/__init__.py
willylee007: Updated `_LOGGER.warning` to `_LOGGER.debug` in bc91771. Thanks for the suggestion.

tests/devices/ac/message_ac_test.py
willylee007: Thanks, agreed that asserting the parsed values for all attributes would improve these tests. To keep this PR focused on `power_saving`, I have retained the new explicit `power_saving` value assertion without expanding the existing attribute checks in this PR.

tests/devices/ac/message_ac_test.py
willylee007: Agreed that parametrization could reduce duplication between similar parser tests. I am keeping this PR scoped to `power_saving`, so I have not included the broader A0/C0 test refactor here.

willylee007:
@chemelli74 The requested logging change has been applied in bc91771, and I have replied to the review threads. The AC tests and full pre-commit checks pass. Could you please take another look when convenient? Thanks.

PR #501, opened 2026-07-14
fix(ac): decode subtype-8 0x7e setpoint and ignore stale C0 temperatures

luoluomeng:
## Summary
- Fix AC subtype-8 setpoint decoding from the `0x7e` new-protocol payload.
- Decode target temperature from byte-1 bitfield (observed on real device traces), with a safe fallback to legacy byte-3 mapping for variant payloads.
- Prevent brief UI flicker by ignoring stale C0 temperature fields once B1/B5 new-protocol temperatures are observed.
- Add regression tests for:
  - `0x7e` temperature parsing (primary mapping + fallback path)
  - stale C0 temperature suppression after new-protocol temperature updates

## Why
Some subtype-8 devices were reporting implausible or sticky setpoints (for example, always `26.0`) even when the AC app/device setpoint changed. Trace analysis showed that for this model family, the effective setpoint is encoded in `0x7e` byte-1 rather than relying only on the previous mapping path.

## Validation
- Live device trace verification:
  - setpoint `25.0` is decoded as `25.0`
  - setpoint `26.5` is decoded as `26.5`
- IDE diagnostics: no errors in modified files.

### DIFF
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -7,6 +7,7 @@ from typing import Any, ClassVar, Unpack
 
 from midealocal.const import DeviceType
 from midealocal.device import MideaDevice, MideaDeviceInitKwargs
+from midealocal.message import ListTypes
 
 from .message import (
     MessageACResponse,
@@ -82,6 +83,13 @@ class DeviceAttributes(StrEnum):
     self_clean = "self_clean"
     pmv = "pmv"
     error_code = "error_code"
+
+
+STALE_C0_TEMPERATURE_ATTRIBUTES = (
+    DeviceAttributes.target_temperature,
+    DeviceAttributes.indoor_temperature,
+    DeviceAttributes.outdoor_temperature,
+)
 
 
 class MideaACDevice(MideaDevice):
@@ -200,6 +208,9 @@ class MideaACDevice(MideaDevice):
         self._customize_max_temperature: float | None = None
         self._power_analysis_method: int = 1
         self._default_power_analysis_method: int = 1
+        # Once 0x7e-derived temperatures are seen, ignore stale C0 temperature
+        # fields to avoid brief UI flicker caused by query ordering.
+        self._prefer_new_protocol_temperature: bool = False
         self.set_customize(customize)
 
     @property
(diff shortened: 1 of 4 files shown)


midealocal/devices/ac/__init__.py
```diff
@@ -198,6 +199,9 @@ def __init__(
         self._customize_max_temperature: float | None = None
         self._power_analysis_method: int = 1
         self._default_power_analysis_method: int = 1
+        # Once 0x7e-derived temperatures are seen, ignore stale C0 temperature
+        # fields to avoid brief UI flicker caused by query ordering.
+        self._prefer_new_protocol_temperature: bool = False
```
chemelli74, line 213: Cannot we just check first 0x7e, if there ignore C0 ?

luoluomeng: C0 is typically received before 0x7E, which can cause the UI to momentarily display an incorrect value before it is corrected.

This change makes the parser prioritize 0x7E. Once 0x7E has been seen, subsequent C0 packets will be ignored.

luoluomeng:
BTW, this PR only fixes the temperature display issue on new devices; it still isn't controllable.

wuwentao:
Resolved conflicts with current main and pushed the merge commit. I also tightened the stale-C0 suppression to key off an explicit parser signal when the subtype-8 0x7e temperature payload is seen, which addresses the review thread about checking for 0x7e directly.

Verification:
- env UV_CACHE_DIR=/private/tmp/midea-local-uv-cache PYTHONPYCACHEPREFIX=/private/tmp/midea-local-pycache uv run --with-requirements requirements.txt --with pytest --with typing_extensions python -m pytest tests/devices/ac/device_ac_test.py tests/devices/ac/message_ac_test.py -q
- env UV_CACHE_DIR=/private/tmp/midea-local-uv-cache PYTHONPYCACHEPREFIX=/private/tmp/midea-local-pycache uv run --with ruff ruff check midealocal/devices/ac/__init__.py midealocal/devices/ac/message.py tests/devices/ac/device_ac_test.py tests/devices/ac/message_ac_test.py

Reviewed and updated by OpenAI Codex (GPT-5).

wuwentao:
Corrected the PR history after the previous merge commit caused CI trouble.

What changed:
- Rewrote the PR branch to a single linear commit on top of current main: bfe02ce fix(ac): decode subtype-8 0x7e temperatures
- Removed the merge commits from the PR history.
- Fixed the dynamic subtype-8 parser flag assertion so mypy is happy.
- Approved and reran the forked PR workflows.

Current verification:
- PR Commit Messages Check: passed
- Python Build and Test matrix for 3.12, 3.13, 3.14 on ubuntu/macos/windows: passed
- Local focused AC tests: 57 passed
- Local ruff check on touched files: passed

Updated by OpenAI Codex (GPT-5).

PR #503, opened 2026-07-18
fix: ignore messages shorter than the header in pre_process_message

vyakunin:
## Problem

`MideaDevice.pre_process_message` indexes `msg[9]` unconditionally:

```python
def pre_process_message(self, msg: bytearray) -> bool:
    """Pre process message."""
    if msg[9] == MessageType.query_appliance:
```

A device that answers with fewer than 10 bytes raises `IndexError`. That
propagates out of `parse_message`, aborting the entire status parse, so every
attribute keeps its previous value and the device looks permanently stale
rather than reporting a parse failure for the one bad frame.

## How I hit it

A 0xFA tower fan (model `56000211`, protocol V3) answers some queries with a
4-byte `01000000` payload. Via `midea_ac_lan`, the `fan` entity was stuck
reporting `off` no matter what the fan was actually doing — turning it on from
the app never updated Home Assistant. Guarding this locally fixed reads
immediately, and the fan has been reporting power/mode/speed/oscillation
correctly since.

## Fix

Return early when the message is too short to contain the header, matching how
other malformed frames are skipped, and log it at debug level. Adds a
`MESSAGE_TYPE_INDEX` constant so the header offset is named rather than a bare
`9` at the use site.

## Test

`tests/device_test.py::test_pre_process_message_short_message` asserts every
length from 0 up to the header offset is ignored. Verified it actually
falsifies: with the guard removed it fails with
`IndexError: bytearray index out of range`, and passes with it.

Note the test is deliberately module-level rather than added to the existing
`MideaDeviceTest` class — that class name doesn't match pytest's default
`python_classes = Test*`, so its cases are silently never collected (the file
currently contributes only `test_fetch_v2_message`). Happy to send that as a
separate PR if you'd like it fixed; I left it alone to keep this one focused.

## Checks

- `python -m pytest ./tests/` → 244 passed (243 before, +1 new)
- `mypy midealocal` → Success, no issues in 87 source files
- `pylint --rcfile=pylintrc midealocal/device.py` → 10.00/10
- `ruff format --check` → clean

`ruff check` reports one `RUF043` in `tests/device_test.py:255`, but that is
pre-existing on an unmodified checkout of `main` (a newer ruff flagging
existing code), so I left it out of scope.


wuwentao:
Updated this PR with a follow-up commit after review.

Changes made:
- Short appliance-query payloads now return `False` from `pre_process_message`, so `parse_message` actually skips malformed short frames instead of forwarding them into device-specific parsers.
- Added a regression test that verifies short decrypted payloads do not call `process_message` and keep `_appliance_query` enabled for the next valid response.
- Cleaned the existing `pytest.raises(..., match=...)` regex in the touched test file so current Ruff passes.

Verification run locally:
- `python -m pytest tests/device_test.py -q` via `uv run`: 3 passed
- `ruff check midealocal/device.py tests/device_test.py`: passed
- `ruff format --check midealocal/device.py tests/device_test.py`: passed
- `git diff --check`: passed

The PR is non-draft and mergeable; awaiting any GitHub checks that get scheduled for the pushed commit.

Reviewed and updated by Codex (GPT-5).

wuwentao:
I will merge it and publish a new release.
if this changes can't work or have bug, we still can fix it via a new PR, thank you very much for your PR.

chemelli74:
@wuwentao this change reverted #514 that again silently skip a lot of tests :-(

PR #507, opened 2026-07-20
feat(ac): add group data

s00500:
- **feat(ac): add group 1, 2 and 7 data queries**


chemelli74:
> * **feat(ac): add group 1, 2 and 7 data queries**
> * **smal bugfix**

Please be so kind to raise 2 different PR, one for the fix and one for the new feature(s).
thank you.

s00500:
Squashed the two commits, second one was needed to fix the tests for the CI, commit message was not too good ;-)

chemelli74:
> Squashed the two commits, second one was needed to fix the tests for the CI, commit message was not too good ;-)

Your PR description says "feat + small bugfix".
What I would like to see is 1 PR for the bugfix and 1 PR for the feat.

Thank you

s00500:
Sure, 
but the bugfix was not a real bugfix, rather related to the main change to fix the tests and ensure checks succeed.

I fixed the pr description now as well

wuwentao (approved):
## Review: `feat(ac): add group data` ✅

Reviewed at head `37a888e` in an isolated worktree, with the full quality gate reproduced locally (pinned to the repo's CI tool versions).

### Summary
Adds AC **group 1 / 2 / 7** data queries and refactors the three existing group queries (`0`, power=`4`, humidity=`5`) onto a shared `MessageGroupDataQuery` base that encodes the group as `0x40 | group`. New engineering/service telemetry exposed:

- **Group 1** — compressor frequency/target/current/voltage + refrigerant circuit temps T1–T4 and discharge pipe (TP)
- **Group 2** — indoor fan speed / target + condensate water-pump state
- **Group 7** — real-time compressor power

### Verification (traced, not assumed)
- ✅ **`__init__.py` "skipped from then on" comment is accurate.** Traced end-to-end: `connect(check_protocol=True)` runs at startup (`device.py:684→244`); a query that times out is appended to `_unsupported_protocol` per **class name** (`device.py:428`), and later polls skip those classes (`device.py:396`). Each group is a distinct class, so an unanswered group is correctly dropped individually, and normal polling doesn't block on it (`recv` is gated by `if check_protocol:`).
- ✅ **Refactor is behaviour-preserving.** `0x40 | group` reproduces the original bodies (group 0→`0x40`, power→`0x44`, humidity→`0x45`); the overridden `body` still omits the `message_id` byte exactly as before. Parametrized test covers all six groups.
- ✅ **Min-length guards** (`15 / 9 / 12`) each equal `(highest index accessed + 1)` — no off-by-one.
- ✅ Read-only telemetry correctly excluded from `set_attribute`, with a dedicated test asserting no set-message is ever built.

### Quality gate (local, matching CI)
- `pytest ./tests/` → **all pass** (AC suite 66 → **67** with the added test below)
- `ruff@0.6.9 check` + `format --check` → clean · `mypy` (strict, incl. changed test) → clean · `pylint` → **10.00/10**
- CI on this PR → **all green**

### Change pushed
- `test(ac): cover truncated group 2 query body` (`37a888e`) — adds the missing short-body test for the C1 `0x42` guard, mirroring the existing group 1 / 7 truncated-body tests. Purely additive test coverage; no source changes.

### Optional follow-ups (non-blocking)
- **Absent-sensor sentinel** (`message.py` `_parse_group_one`): raw bytes are decoded directly, so a `0xFF` placeholder for an unpopulated sensor would surface as `112.5 °C` / `255`. This file already treats `0xFF` as `None` in `parse_temperature` and guards humidity with `!= 0`. Worth aligning **if** hardware confirms a sentinel — better as a follow-up than a blind guess, since it's engineering data ported 1:1 from upstream.
- Minor: `discharge_pipe_temperature` is an `int` while sibling temps are `float` (documented as "reported directly in degrees" — fine).

**Assessment: approve.** Clean, well-tested, house-style conforming (named constants, `StrEnum` attributes, tests mirror source), and the refactor tidies the existing group queries. Nice work — thanks to @s00500, @TurboLed and @its-tom for the protocol research.

<sub>🤖 Reviewed & verified by <b>Claude Code</b> (model: <b>Claude Opus 4.8</b>) · tests/ruff/mypy/pylint run locally + CI green</sub>

s00500:
Awesome ! Thanks a lot!

Issue #506, opened 2026-07-19
AI discovered bugs


chemelli74:
I wrote some tests myself and some with the help of AI.

Then I did a full run wth Fable and found a few issues in the code base:

1. **Coverage:** The only remaining uncovered line is `midealocal/devices/dc/__init__.py:160` (an incorrect comparison between `int` and `list[str]`). This is genuine dead production code, not a missing test.

2. **Silent Tests Restored:** The `MideaDeviceTest` test class has been correctly renamed to `TestMideaDevice*`, allowing pytest to discover and execute these ~200 tests that were previously being silently ignored.

3. **Production Bugs Identified** *(outside the scope of this PR, but worth addressing separately)*:
   - `ce/__init__.py`: typo in `"aux_heating'"` (stray trailing single quote in the attribute name).
   - `e6/__init__.py`: mismatch between `min_temperature` and `temperature_min`.
   - `e2/__init__.py`: trailing comma creates a tuple that always evaluates as `True`, breaking the subtype check.
   - `cd/message.py:451`: potential `IndexError` due to an incorrect length guard.
   - `x26/__init__.py`: unhandled `AttributeError` when processing unexpected message bodies.
   - `discover.py:245,320`: deprecated use of `if not child:` on XML `Element` objects. This will become a Python compatibility issue in the future and should be replaced with `if child is None:`.

I'll document them here before I forget, corrections will come

PR #509, opened 2026-07-20
fix(cd): correct length body

chemelli74:
Correct length protection: len(body) > OLD BODY LENGTH (29) → len(body) > 34, aligned to the index actually read (body[34])

Fix 1 issue of #506


wuwentao:
this should not a bug, it's a by design issue.
if packet length >28, it should works with new protocol, this length check also available from midea official lua protocol.
anyway, we can accept this changes and merge it

wuwentao (approved):

PR #384, opened 2025-07-20
feat(a1): remove inaccurate attributes

SoCuul:
## Changes
- Removes `child_lock` as it does not exist on dehumidifiers
- Removes`swing` as it does not exist on dehumidifiers
- Removes `anion` as it does not exist on dehumidifiers

- Adjusts `_modes` to only include the possible dehumidifying odes: “Manual”, “Continuous” & “Max”
- Adjusts `_speeds` to only include the possible dehumidifier fan speeds: “Low” & “High”

---

I've tested all these changes with my personal Midea Cube Dehumidifier, and they work perfectly. The dehumidifier midea-local device now functions properly, and provides the same controls that the Midea app does. There are no longer options that have no effect on the functionality/operation of the dehumidifier.

This will go along with another pull request to the `midea_ac_lan` repository, to remove the references to these attributes that were removed.

To back up my claims that no Midea dehumifiers support these features, here are the owner's manuals for all three dehumidifiers currently being sold by Midea:
- https://www.midea.com/content/dam/midea-aem/us/air-conditioners/dehumidifiers/easydry-22-pint-dehumidifier-mad22c1aws/User%20Manual%20EasyDry%20Dehum%20MADC1YWS.pdf
- https://www.midea.com/content/dam/midea-aem/us/air-conditioners/dehumidifiers/mad20s1qwt/MAD20S1QWT-2024-Cube-Manual.pdf
- https://www.midea.com/content/dam/midea-aem/us/air-conditioners/dehumidifiers/mad50ps1qwt/MAD50PS1QWT-2024-Cube-Manual.pdf

PR #511, opened 2026-07-20
fix(e6): renamed temperature wrong attributes

chemelli74:
Renamed the min_temperature/max_temperature → temperature_min/temperature_max body fields to align with the values ​​in the DeviceAttributes enumeration.

Fix 1 issue of #506

### DIFF
--- a/midealocal/devices/e6/__init__.py
+++ b/midealocal/devices/e6/__init__.py
@@ -20,8 +20,8 @@ class DeviceAttributes(StrEnum):
     heating_power = "heating_power"
     heating_working = "heating_working"
     bathing_working = "bathing_working"
-    min_temperature = "temperature_min"
-    max_temperature = "temperature_max"
+    temperature_min = "temperature_min"
+    temperature_max = "temperature_max"
     heating_temperature = "heating_temperature"
     bathing_temperature = "bathing_temperature"
     heating_leaving_temperature = "heating_leaving_temperature"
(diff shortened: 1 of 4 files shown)


wuwentao (changes requested):
the root cause should be __init__.py defined
```
    min_temperature = "temperature_min"
    max_temperature = "temperature_max"
```
we should change from the source, set the name to the same as value. as this rule exist in most of the device attrs

chemelli74:
Man, I see your advice but still you should give me the time to read and react.

Is not fair to commit on my PR without a previous agreement with the author.

wuwentao (dismissed):
Addressed the requested change by renaming the E6 DeviceAttributes members to temperature_min/temperature_max so they match the exposed attribute values and parsed response fields.

Added focused E6 device coverage for initial attributes and process_message mapping.

Verification:
- ruff check midealocal/devices/e6/__init__.py midealocal/devices/e6/message.py tests/devices/e6/device_e6_test.py
- pytest tests/devices/e6/device_e6_test.py -q
- GitHub PR check matrix passed on Python 3.12, 3.13, and 3.14 across ubuntu-latest, macos-latest, and windows-latest

Reviewed and updated by OpenAI Codex (GPT-5).

wuwentao:
> Man, I see your advice but still you should give me the time to read and react.
> 
> Is not fair to commit on my PR without a previous agreement with the author.

I can understand it, and we already sync it, you can check with old PRs, not 1-3 PRs, and we should have enough time to post a simple reply.
you can submit more code changes and PRs to fix issue these weeks, but you don't have any time to process any exist PR review comments, this is the root cause.
so I don't want to waste time with it, and I also cost more times to resove more PR conflicts.
once it done, I will publish a new release.

anyway,  we still can improve it with new PR,  as there is no any blocker issue or mandatory rule exist for you.

for example: 
PR https://github.com/midea-lan/midea-local/pull/384, it's dead, why?
then, the same PR exist,  https://github.com/midea-lan/midea-local/pull/476, I don't want got the same result again.
so I want to got a quickly response and feedback with contirbutors, NOT always post requements and deny any PR.

anyway, this is really a very small project, code quality is another topic, if there is no any major issue, we can try to help them improve the minor issue or accept it or submit new PR, I have post this info too many times.

PR #527, opened 2026-07-24
fix(cd): don't let SET echo clobber power state (draft, needs device validation)

wuwentao:
> **Draft — needs validation on real CD hardware before merge.** See "Validation needed" below.

## Problem

On CD water heaters, a pure `target_temperature` (or `mode`) SET can silently switch the unit **OFF**.

Every power/mode/temperature write goes out as a single `controlType=0x01` frame that bundles **all three** fields (`midealocal/devices/cd/__init__.py`):

```python
current_power = self._attributes.get(DeviceAttributes.power, False)   # ~L404
...
message.power = current_power                                          # ~L411
...
elif attr == DeviceAttributes.target_temperature:
    message.target_temperature = float(value)                         # never touches message.power
```

For a temperature-only change, `message.power` is whatever `_attributes[power]` currently holds. If that stored value is wrongly `False`, the setpoint change commands the device off.

## Root cause

`process_message` accepts the `power` bit from **every** frame via the generic store path, with **no SET-echo guard** — unlike `mode`, which is already guarded (only recognised keys accepted). SET echoes are known to carry junk in this device (see the many `openPTC`/`Tr`/mode-echo notes in the file). A spurious `power=0` echo overwrites a good ON reading, and that stale OFF is then replayed as an OFF command on the next temperature/mode write.

## Fix

Distrust the `power` bit from SET echoes (`message_type == MessageType.set`), exactly as `mode` is already distrusted. Genuine `status`/`notify`/`query` frames still update `power` normally. Unknown/mocked messages keep prior behaviour (safe for existing tests).

- `+8` lines in `process_message` (compute `is_set_echo`, skip power overwrite on echo)
- `+2` regression tests: echo must not clobber ON; genuine status frame still updates power

## Why here and not in Home Assistant

The downstream integration PR wuwentao/midea_ac_lan#768 worked around this at the entity layer by forcing `power=True` before every setpoint write. That (a) makes it impossible to set a target temperature while the heater is intentionally OFF, and (b) emits two control frames. Fixing it here keeps the entity layer honest.

## Test / lint

```
pytest tests/devices/cd/    ->  65 passed
ruff check (changed files)  ->  All checks passed
ruff format --check         ->  already formatted
mypy midealocal/devices/cd  ->  Success: no issues found
```

## Validation needed before merge

I don't have CD hardware, so please confirm on a real unit + debug log:

1. **SET echoes actually arrive as `message_type == MessageType.set`** (0x02). If a CD SET echo is instead delivered as a notify frame, the guard needs to key off a different discriminator.
2. **After a temperature-only change, a genuine status/notify frame follows** so `_attributes[power]` reconciles quickly (the guard intentionally does not optimistically flip power on a SET echo).
3. Whether the **pre-first-status** default (`power=False` at init, ~L117) also needs handling — deliberately left out of this PR to keep scope tight; can add a follow-up if a temp-set can realistically precede the first status frame.

Refs: wuwentao/midea_ac_lan#768

### DIFF
--- a/midealocal/devices/cd/__init__.py
+++ b/midealocal/devices/cd/__init__.py
@@ -7,6 +7,7 @@ from typing import Any, ClassVar, Unpack
 
 from midealocal.const import DeviceType
 from midealocal.device import MideaDevice, MideaDeviceInitKwargs
+from midealocal.message import MessageType
 
 from .message import (
     MessageCDBase,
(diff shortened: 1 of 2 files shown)


chemelli74:
Superseded by #776

PR #481, opened 2026-06-25
feat(ac): decode B5 capabilities into feature flags

Pulpyyyy:
## Summary
Decode the AC **B5 capability** values into boolean feature flags, exposed via a new `MideaACDevice.capabilities` property. The B5 response is already queried (`MessageCapabilitiesQuery` in `build_query`) and parsed by `XB5MessageBody` — this PR adds the *interpretation* of the values.

## Why
The raw B5 byte of each capability is **not** a 0/1 flag — each one has its own value semantics (reverse-engineered, matching the `msmart` project). Decoding them lets a consumer know what the device really supports:
- modes: `heat` / `cool` / `dry` / `auto`
- swing: horizontal / vertical
- fan: silent / low / medium / high / auto / custom
- `eco`, `anion`, `turbo` (cool/heat), `display`

## Non-regression (important)
- **Purely additive**: a new `capabilities` dict (populated only from B5) + `MideaACDevice.capabilities`. No existing parsing/attribute is changed.
- Devices that don't report B5 → `capabilities == {}` → consumers fall back to their defaults. **No behaviour change for existing devices.**
- `mypy` (strict) passes; lines ≤ 88.

## Verified on real hardware
Tested on a cooling-only portable AC (model `00000Q17`). The decode reconciles **9/9** with what the unit actually accepts/rejects (cool/dry/fan, low/high/auto fan, no swing, no eco/turbo, 17–30 °C). Decrypted Lua plugin used as reference: https://gist.github.com/Pulpyyyy/02d3d42d62ab17c10ab3f799b00fdf61

## Companion PR
`wuwentao/midea_ac_lan` consumes `capabilities` to align the climate entity (hvac_modes / fan_modes / swing / presets) — opened separately: wuwentao/midea_ac_lan#862.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #564, opened 2026-07-24
docs: add E2 and AC (0xAC) reference lua plugins

Pulpyyyy:
## What

Adds two decompiled Midea device plugins to the `lua/` reference corpus:

- `T_0000_E2_24.lua` — E2 electric water heater (subtype 24)
- `T_0000_AC_00000Q17_2024013001.lua` — 0xAC air conditioner, cooling-only
  portable unit (model 00000Q17)

## Why

The `lua/` folder currently ships B6 / CD / ED plugins but has no E2 or 0xAC
reference. These two are the protocol source I used to reverse-engineer and
verify the recent E2 and AC work:

- #477 — E2 `memory` (Memo U)
- #478 — AC per-mode min/max temperature (B5)
- #481 — AC B5 capabilities decode

Adding them here keeps the reference corpus complete and lets others cross-check
the 0xE2 / 0xAC decode against the original plugin.

## Notes

Both files were obtained from the Midea cloud (`download_lua`) for my own
devices and are committed verbatim (no edits). `codespell` and
`detect-private-key` pass on them.


chemelli74:
Thank you so much @Pulpyyyy !

As soon as CI passes, I'll merge this PR.

PR #563, opened 2026-07-24
fix(ac): reject invalid subtype-8 temperatures

Sanjays2402:
Fixes wuwentao/midea_ac_lan#893

Subtype-8 `0x7e` payloads currently become authoritative before their length or decoded values are validated. An unrelated payload can therefore decode the indoor temperature as `-25.0 C` and suppress the valid `C0` room-temperature updates that follow.

This only marks a payload authoritative after both temperatures pass validation, with regressions for the reported sentinel value and truncated payloads. All 125 AC device/message tests pass.

PR #567, opened 2026-07-26
fix(ac): only decode 0x7e temperatures on subtype-8 devices

### LINKED ISSUE #893: Bug(AC): current_temperature shows -25°C after v0.7.0 update, though raw indoor_temperature is correctly parsed as 25.8°C
### Verify Issue Submission

- [x] I have read the [README documentation](https://github.com/wuwentao/midea_ac_lan/blob/main/README.md) , understand how to use midea_ac_lan and enable debug log.
- [x] I have read the [device documentation](https://github.com/wuwentao/midea_ac_lan/blob/main/README.md#4-supported-appliances) for my device in the README, but the problem still can't fix.
- [x] I have submitted the device type and SN information.
- [x] I have uploaded the full debug log file, rather than just paste a parts of log based on my own judgment.
- [x] I have accurately and clearly described what the current error is and what the expected correct result should be.
- [x] I have searched for my issue in the [Issue Tracker](https://github.com/wuwentao/midea_ac_lan/issues) and have not found it.

### HA core version

2026.7.3

0.7.0 (midea-local 6.11.0)

0.6.12

### Device type and Device SN

Air Conditioner, model 00000Q18 (8), type 172 (0xAC), SN 000000P0000000Q188F2BDD15F4A0000

### Working mobile app

Not applicable - device is controlled solely via Home Assistant (midea_ac_lan local integration), no mobile app used day-to-day

### The detail description of bug/problem

Device: Air Conditioner, model 00000Q18 (8), type 172 (0xAC/DeviceType.AC), one device only.

Error: After updating midea_ac_lan from v0.6.12 to v0.7.0 today (2026-07-24) and restarting Home Assistant, the climate entity's "Currently" (current_temperature) reads a constant -25 C / 40% and does not change, regardless of actual room conditions. This is visible directly on the device page in HA. The unit worked correctly on v0.6.12.

Expected: current_temperature should reflect the actual indoor temperature, as it did before the update (previously matched real room temperature, e.g. ~25-26 C in Cool mode).

Proof from the attached debug log: the raw protocol messages received from the device (ListTypes.C0 query responses) correctly parse indoor_temperature as 25.8 the entire time (see lines with "Received: {... 'indoor_temperature': 25.8 ...}" e.g. at 17:09:20.346, 17:10:00.292, 17:10:30.323). So the device itself is reporting a normal, correct indoor temperature over the wire. However, the climate entity's displayed current_temperature in the HA UI is stuck at -25 C, not 25.8 C. This points to a regression in how the climate entity's current_temperature attribute is now being derived/mapped in v0.7.0 (possibly related to the "expose model-reported diagnostics and airflow" or "derive climate modes/fan/swing/presets from B5 capabilities" changes in this release), rather than a device/communication problem.

Rolling back to v0.6.12 is expected to resolve this (not yet tested, but the value read correctly before the update).

### The logs

[debug.rtf](https://github.com/user-attachments/files/30355105/debug.rtf)

1. DEBUG log MUST be enabled (Please refer to README document), Submit issue without debug log will be directly closed
2. Execute the action to trigger error log（you should description the error action）
3. Upload the debug log file as attachment。
4. Summary: Device SN describes detailed device information. Please clearly describe what the problem and error are. What should the correct expected result be? Finally: How can you prove what you're saying is correct? The debug log file is used to prove the error genuinely exists, and app screenshots or other data are used to prove the correct expected result.
5. Please don't paste the log content! upload it as a attachment file.
6. Please don't paste the log content! upload it as a attachment file.
7. Please don't paste the log content! upload it as a attachment file.
...

caibinqing:
## Summary

The `0x7e` new-protocol temperature payload is decoded for **every** AC that reports the tag, even though the offsets were reverse-engineered from subtype-8 traces (model 22013279). Other subtypes send tag `0x7e` with unrelated content, and the same offsets then decode to a bogus room temperature that permanently suppresses the valid `C0` temperatures.

This gates the decoding on the device subtype.

## Reproduction

Home Assistant `midea_ac_lan` 0.7.0 (`midea-local==6.11.0`), AC model **22251759**, subtype **32773**, protocol 3. The room temperature sensor sits at **-15.0 °C** and the outdoor temperature entity has no value at all, while the AC itself is fine.

Debug capture from the device — the same query round trip, two frames:

```
B1: b1017e000037a05b40667f7f003f004800000000000000000000000400200000
    000000000000008028281e040000640800002400200000000000000000001f
    -> has_subtype8_temperature: True, target 25.5, indoor -15.0, outdoor None

C0: c00159667f7f003f00080064740d005600000080800000005802000037ffcaf8
    -> target 25.5, indoor 25.6, outdoor 33.5
```

The `C0` frame is correct (25.6 °C indoor, 33.5 °C outdoor, matching the T4 sensor from the `C1` group query). It is discarded.

In the `0x7e` payload of that B1 frame, `data[39] = 0x00` and `data[40] = 0x64`:

```
(0x00 - 50) / 2 + 0x64 * 0.1 = -25 + 10.0 = -15.0
```

`_parse_subtype8_temperatures` also forces `outdoor_temperature = None`, and `process_message` latches `_prefer_new_protocol_temperature`, so from the first B1 frame onwards every `C0` frame has its target/indoor/outdoor temperature skipped for the rest of the session.

## Why the range check isn't enough

#563 rejects a payload whose decoded temperatures fall outside 10–40 °C, which does hide this particular device's symptom. But it is a heuristic over bytes that mean something else entirely on other models: any unrelated payload that happens to decode into 10–40 °C still latches `_prefer_new_protocol_temperature` and permanently suppresses valid `C0` temperatures — this time with a plausible-looking wrong value, which is harder to notice and to report.

## Changes

- `MessageACResponse` / `XBXMessageBody` take a `subtype8_temperature` flag (default `False`), threaded the same way as the existing `power_analysis_method`; the `0x7e` tag is only decoded when it is set.
- `MideaACDevice` sets it from `subtype == 8`.
- The range validation from #563 is kept as a second line of defence for subtype-8 devices.

Gating on the subtype alone rather than an exact `(model, subtype)` pair in `AC_MODEL_CAPABILITIES` keeps every subtype-8 device working, including models other than 22013279 — #501 mentions "some subtype-8 devices". Happy to switch to the exact-pair mechanism if you prefer that consistency.

## Tests

- The four existing `0x7e` cases now pass `subtype8_temperature=True`.
- New message-level case: a payload decoding to **in-range** 26.0 / 28.8 °C is still ignored when the device isn't subtype 8 — the case the range check cannot catch.
- New device-level case: the same B1 frame is adopted by a subtype-8 device (and latches) but ignored by model 22251759 / subtype 32773, which keeps its `C0` temperatures.

`pytest`: 1452 passed. `mypy` and `pylint` (10.00/10) clean; `ruff check` / `ruff format` unchanged from the pre-change baseline.

Fixes wuwentao/midea_ac_lan#893

### DIFF
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -159,6 +159,13 @@ AC_MODEL_CAPABILITIES = {
     ),
 }
 
+# The 0x7e new-protocol temperature payload was decoded from subtype-8 traces
+# (model 22013279). Other subtypes emit tag 0x7e as well, but with unrelated
+# content: on model 22251759 / subtype 32773 the same offsets decode to
+# -15.0 C and permanently suppress the correct C0 temperatures.
+# https://github.com/wuwentao/midea_ac_lan/issues/893
+SUBTYPE8 = 8
+
 STALE_C0_TEMPERATURE_ATTRIBUTES = (
     DeviceAttributes.target_temperature,
     DeviceAttributes.indoor_temperature,
@@ -318,6 +325,8 @@ class MideaACDevice(MideaDevice):
         self._customize_max_temperature: float | None = None
         self._power_analysis_method: int = 1
         self._default_power_analysis_method: int = 1
+        # Only subtype-8 devices carry temperatures in the 0x7e payload.
+        self._subtype8_temperature: bool = int(self.subtype) == SUBTYPE8
         # Once 0x7e-derived temperatures are seen, ignore stale C0 temperature
         # fields to avoid brief UI flicker caused by query ordering.
         self._prefer_new_protocol_temperature: bool = False
@@ -386,7 +395,11 @@ class MideaACDevice(MideaDevice):
 
     def process_message(self, msg: bytes) -> dict[str, Any]:  # noqa: C901
         """Midea AC device process message."""
-        message = MessageACResponse(bytearray(msg), self._power_analysis_method)
+        message = MessageACResponse(
+            bytearray(msg),
+            self._power_analysis_method,
+            self._subtype8_temperature,
+        )
         _LOGGER.debug("[%s] Received: %s", self.device_id, message)
         new_status = {}
         has_fresh_air = False
--- a/midealocal/devices/ac/message.py
+++ b/midealocal/devices/ac/message.py
@@ -1042,7 +1042,12 @@ class XA1MessageBody(XMessageBody):
 class XBXMessageBody(NewProtocolMessageBody):
     """AC BX message body. body[0] b0/b1, body[1] propertyNumber, cursor 2."""
 
-    def __init__(self, body: bytearray, bt: int) -> None:
+    def __init__(
+        self,
+        body: bytearray,
+        bt: int,
+        subtype8_temperature: bool = False,
+    ) -> None:
         """Initialize AC BX message body."""
         super().__init__(body, bt)
 
@@ -1089,8 +1094,10 @@ class XBXMessageBody(NewProtocolMessageBody):
                 len(data) > SELF_CLEAN_ACTIVE_STATUS_BYTE
                 and data[SELF_CLEAN_ACTIVE_STATUS_BYTE] != 0
             )
-        if SUBTYPE8_TEMPERATURE_TAG in params and self._parse_subtype8_temperatures(
-            params[SUBTYPE8_TEMPERATURE_TAG],
+        if (
+            subtype8_temperature
+            and SUBTYPE8_TEMPERATURE_TAG in params
+            and self._parse_subtype8_temperatures(params[SUBTYPE8_TEMPERATURE_TAG])
         ):
             self.has_subtype8_temperature = True
 
@@ -1102,6 +1109,10 @@ class XBXMessageBody(NewProtocolMessageBody):
         setpoint in byte 1:
         - low 6 bits encode 0.5C steps with a +11.5C offset
         - bit 0x40 adds an extra +0.5C
+
+        Only call this for subtype-8 devices: other subtypes send tag 0x7e with
+        unrelated content, where these offsets decode to a bogus temperature
+        that would then suppress their valid C0 temperatures.
         """
         if len(data) <= SUBTYPE8_TEMPERATURE_MIN_LENGTH:
             return False
@@ -1542,7 +1553,12 @@ class XBBMessageBody(MessageBody):
 class MessageACResponse(MessageResponse):
     """AC message response."""
 
-    def __init__(self, message: bytearray, power_analysis_method: int = 3) -> None:
+    def __init__(
+        self,
+        message: bytearray,
+        power_analysis_method: int = 3,
+        subtype8_temperature: bool = False,
+    ) -> None:
         """Initialize AC message response."""
         super().__init__(message)
         # dataType 0x05 and messageBytes[0] 0xA0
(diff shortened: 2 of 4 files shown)


Amejonah1200:
Apparently https://github.com/midea-lan/midea-local/commit/0953364556d813b99bbb6b2651913b3a17a7a944 is already on main branch

caibinqing:
[`0953364`](https://github.com/midea-lan/midea-local/commit/0953364556d813b99bbb6b2651913b3a17a7a944) (#563) is the base commit of this branch — this PR sits on top of it and keeps its range check as-is.

That check validates the decoded *values*, not the premise for decoding them. On non-subtype-8 devices those bytes mean something else entirely, so a payload that happens to decode into 10–40 °C is still accepted, still latches `_prefer_new_protocol_temperature`, and still permanently suppresses the valid `C0` temperatures — now with a believable wrong value and no outdoor temperature at all. The new test covers exactly that case.

Concretely: `0953364` does fix my own device (subtype 32773), but only because its bytes 39/40 happen to decode to −15 °C and fall outside the range. Byte 39 landing anywhere in `0x46`–`0x82` would latch again, silently. The `0x7e` offsets came from subtype-8 traces, so this gates the decode on `subtype == 8` and keeps that range check as a second line of defence for the devices that really do use the tag.

wuwentao (approved):
## Review: approved ✅

Reviewed at head `7dad7a0`. This correctly fixes the root cause behind wuwentao/midea_ac_lan#893: the `0x7e` new-protocol offsets were reverse-engineered from subtype-8 traces, so decoding them on other subtypes produces a bogus room temperature that latches `_prefer_new_protocol_temperature` and permanently suppresses the valid `C0` temperatures. Gating the decode on `subtype == 8` addresses the premise, not just the symptom, while keeping #563's range check as a second line of defence.

### What I checked
- **Design** — `subtype8_temperature` flag (default `False`) is threaded through `MessageACResponse` → `XBXMessageBody` exactly like the existing `power_analysis_method`, and only the `SUBTYPE8_TEMPERATURE_TAG` (0x7e) block is gated. Every other new-protocol tag (indirect_wind, indoor_humidity, breezeless, screen_display, fresh_air, …) stays ungated. Both production call sites pass the flag explicitly; the default is safe.
- **Correctness** — the `and` short-circuit means `_parse_subtype8_temperatures` is never invoked when the flag is `False`. `int(self.subtype) == SUBTYPE8` matches the existing `int(self.subtype)` idiom used for `_model_key`.
- **Tests** — verified the two new tests are non-vacuous: the built frames carry a `0x7e` payload that decodes to in-range 26.0 / 28.8 °C, so they *would* be adopted if the gate were open. The subtype-8 branch asserts adoption + latch; the non-subtype-8 branch asserts the temps are ignored and the latch stays clear — precisely the case the range check alone cannot catch.

### Local CI reproduction (pinned to CI versions, isolated worktree)
| check | result |
| --- | --- |
| `pytest` | **1452 passed** |
| `ruff@0.6.9 check` / `format --check` | clean |
| `mypy` (repo scope) | clean |
| `pylint --ignore-paths=tests` | clean (exit 0) |
| coverage | `ac/__init__.py` 100%, `ac/message.py` 99% |

### Notes (non-blocking)
- The single uncovered line in `message.py` (`return False` in the fallback out-of-range branch) is **pre-existing from #563**, not introduced here — confirmed via `git blame` and a base-commit coverage run. Optional follow-up: a subtype-8 payload where both the primary setpoint (`data[1]=0x3F` → 43.0) and the legacy fallback (`data[3]=0x00` → −25.0) are out of range would close it.
- On the `subtype == 8` vs exact `(model, subtype)` question you raised: gating on the subtype is the right call — #501 references multiple subtype-8 models, and an exact-pair gate would regress those.

Approving and ready to merge.

---
🤖 Reviewed by **Claude Code** (model: Claude Opus 4.8) via the `github-pr-improve` workflow.

wuwentao:
@luoluomeng as there is more PR exist in midea_ac_lan side, so I have merged this PR and will try to publish a new release to fix bug in midea_ac_lan side.
could you help to test/review and confirm the fix with your device, thanks

in addition, we can accept bugs with new feature support, we can continue improve it with new PRs.
thanks.

luoluomeng:
Sorry, my mistake.
It seems I tested a different version of the code than the one I submitted. Sorry for the confusion.
The correct dataset is data[40]/data[41], not data[39]/data[40].
Apologies again.

PR #569, opened 2026-07-27
fix(b0): assemble cloudmenuid as a big-endian 3-byte value

kreynolds:
## Problem

`B0Message31Body` assembles the cloud menu id like this:

```python
self.cloudmenuid = body[2] * (16 ^ 4) + body[3] * (16 ^ 2) + body[4]
```

In Python `^` is bitwise XOR, not exponentiation, so `16 ^ 4` is `20` and `16 ^ 2` is `18`
— not the intended `65536` and `256`. The result is a wrong value for any response whose
menu id needs more than one byte:

| `body[2:5]` | current | expected |
| --- | --- | --- |
| `00 00 8D` | 141 | 141 |
| `01 02 03` | **59** | 66051 |

Devices that only ever report a single-byte menu id are unaffected, because `body[2]` and
`body[3]` are zero for them and the two wrong multipliers get multiplied by zero. That is
probably why this has gone unnoticed.

## Fix

Use `int.from_bytes(body[2:5], "big")`, which matches the idiom already used in
`discover.py` and `security.py` and removes the magic numbers entirely.

## Tests

Two cases added to `TestB0MessageBodies`:

- `test_31_body_cloudmenuid_single_byte` — `00 00 8D` → `0x8D`. Passes before and after,
  documenting the case that already worked.
- `test_31_body_cloudmenuid_multi_byte` — `01 02 03` → `0x010203`. Fails on `main` with
  `assert 59 == 66051`, passes with the fix.

Full suite is green locally (1454 passed) and `pre-commit run --all-files` is clean.

## How this was found

While decoding B0 traffic from a Toshiba ML-EM34P(SS) microwave (device type `0xB0`,
`midea_b0_1434`), where the menu id distinguishes presets — `0x8D` popcorn, `0x8E` potato,
`0x8F` sensor reheat. That appliance reports single-byte ids, so it is **not** affected by
this bug; the defect was spotted by reading the parser rather than from a broken reading.
I have no device that reports a multi-byte menu id, so the corrected value is verified by
construction and by the intent of the expression, not against hardware.


wuwentao (approved):
## Review: approved ✅

Verified in an isolated worktree against the CI-pinned toolchain.

**Bug confirmed.** In Python `^` is bitwise XOR, so `16 ^ 4` → `20` and `16 ^ 2` → `18`, not the intended `65536`/`256`. This is the only occurrence of the XOR-as-power pattern in the codebase (`grep -rE '[0-9]+ \^ [0-9]+' midealocal/`).

**Fix is correct and idiomatic.** `int.from_bytes(body[2:5], "big")` matches the existing convention in `security.py:341` and `discover.py:211`, and removes the magic numbers per the repo's "no magic numbers" rule. The `body` length guard (`> MIN_MSG_BODY`, i.e. ≥ 16 bytes) means the `[2:5]` slice is always safe.

**Tests are genuine regression tests.**
- `test_31_body_cloudmenuid_multi_byte` fails on `main` with exactly `assert 59 == 66051`, passes with the fix.
- `test_31_body_cloudmenuid_single_byte` documents the previously-working single-byte case.

**Checks run (worktree, CI-pinned):**
- `pytest tests/devices/b0/` → 33 passed
- `ruff@0.6.9 check` + `format --check` → clean
- `mypy` → no issues

**Non-blocking note:** `cloudmenuid` is currently only ever assigned, never read elsewhere in the codebase — so this bug has no runtime impact today. The fix is still correct and future-proofs the parser. No changes requested.

---
🤖 Reviewed by Claude (Opus 4.8) via Claude Code

chemelli74:
> **Non-blocking note:** `cloudmenuid` is currently only ever assigned, never read elsewhere in the codebase — so this bug has no runtime impact today. The fix is still correct and future-proofs the parser. No changes requested.

Let' drop it if unused please.
Clean code is an important goal.

wuwentao:
> > **Non-blocking note:** `cloudmenuid` is currently only ever assigned, never read elsewhere in the codebase — so this bug has no runtime impact today. The fix is still correct and future-proofs the parser. No changes requested.
> 
> Let' drop it if unused please. Clean code is an important goal.

No, all the codes/feature should works well, as there is no device to verify the real test result(should support this feature).
and user can easy to got the result via debug log,  also can add new features with a device test log and confirm the value.

we just keep it in debug log output and not publish it via HA integration now. 
if we really want to support it,  just add a new attr name, but we should confirm the text with a real device, as raw output is hard to read/understand for end user.

chemelli74:
> > > **Non-blocking note:** `cloudmenuid` is currently only ever assigned, never read elsewhere in the codebase — so this bug has no runtime impact today. The fix is still correct and future-proofs the parser. No changes requested.
> > 
> > 
> > Let' drop it if unused please. Clean code is an important goal.
> 
> No, all the codes/feature should works well, as there is no device to verify the real test result(should support this feature). and user can easy to got the result via debug log, also can add new features with a device test log and confirm the value.
> 
> we just keep it in debug log output and not publish it via HA integration now. if we really want to support it, just add a new attr name, but we should confirm the text with a real device, as raw output is hard to read/understand for end user.

@wuwentao, currently the code as:

```python
self.cloudmenuid = body[2] * (16 ^ 4) + body[3] * (16 ^ 2) + body[4]
```

but is never used, so what's the meaning in having it ?
Please let me understand why you want to keep it

PR #570, opened 2026-07-27
feat(b0): map mode 0xA1 time defrost in _mode31

kreynolds:
## Problem

Subtype-2 microwaves report mode `0xA1` while a **time-based defrost** runs. `_mode31` has
no entry for it, so `DeviceAttributes.mode` resolves to `None` for the entire cycle.

Observed on a **Toshiba ML-EM34P(SS)** — device type `0xB0`, model name `midea_b0_1434`,
SN `0000B03470EM34A2E06101101434PVJ0`, subtype 2. Starting a one-minute time defrost from
the appliance panel produces this in `midealocal` debug logging:

```
Received:      ... 'mode': 161, 'fire_power': 3 ...
Status update: {'status': 'Working', 'time_remaining': 60,
                'mode': None, 'fire_power': 'Medium Low', ...}
```

Raw mode `161` (`0xA1`) → `None`.

Every other mode this appliance uses already resolves correctly through `_mode31`, which is
why this is the only gap:

| raw | resolves to | panel action |
| --- | --- | --- |
| `0x01` | `Microwave` | manual cook |
| `0xA0` | `Unfreeze` | defrost **by weight** |
| **`0xA1`** | **`None`** | defrost **by time** |
| `0xE0` | `Auto` | menu presets (popcorn, potato) |
| `0xE2` | `Humidit Auto` | sensor reheat |

`_fire_power31` also resolves every value this appliance emits (`3` Medium Low, `5` Medium,
`8` Midium High, `10` High), so nothing else needed changing.

## Fix

One entry in `_mode31`:

```python
0xA1: "Time Unfreeze",
```

Named to sit alongside the existing `0xA0: "Unfreeze"`, since the two are the appliance's
weight-based and time-based defrost respectively. **If your protocol source has a canonical
name for `0xA1`, please use that instead** — I inferred this one from panel behaviour.

## Tests

`test_process_message_31_time_unfreeze` drives a `0x31` body with `mode = 0xA1` and
`fire_power = 0x03`, mirroring the observed frame. It fails on `main` with
`assert None == 'Time Unfreeze'` and passes with the change.

Full suite green locally (1453 passed); `pre-commit` clean.

## Scope

This is one appliance and one subtype. I have not assumed anything about subtype-0 devices
or about `_mode`, and have deliberately not touched either. Happy to supply the full debug
log for the cycle if useful.


wuwentao (approved):
## Review: approved ✅

Verified in an isolated worktree against the CI-pinned toolchain.

**Gap confirmed.** `0xA1` is absent from `_mode31`, so `DeviceAttributes.mode` resolves to `None` for the whole time-defrost cycle on subtype-2 devices. Every other mode this appliance emits already resolves, so this is the only missing entry.

**Fix is minimal and correctly scoped.** The single `0xA1: "Time Unfreeze"` entry is added to `_mode31` only (subtype > 0), sitting alongside `0xA0: "Unfreeze"`; `_mode` and subtype-0 behaviour are untouched. Adding a dict key is non-breaking. Naming follows the existing convention.

**Test is a genuine regression test.** `test_process_message_31_time_unfreeze` fails on `main` with exactly `assert None == 'Time Unfreeze'`, passes with the fix. It drives the full `process_message` path with the observed frame (`mode=0xA1`, `fire_power=0x03`).

**Checks run (worktree, CI-pinned):**
- `pytest tests/devices/b0/` → 32 passed
- `ruff@0.6.9 check` + `format --check` → clean
- `mypy` → no issues

**Non-blocking note:** the `"Time Unfreeze"` name is inferred from panel behaviour, as you flagged. If the protocol source has a canonical name for `0xA1`, worth swapping to it — but that's a maintainer call and doesn't block merge.

---
🤖 Reviewed by Claude (Opus 4.8) via Claude Code

PR #572, opened 2026-07-28
fix(ac): model 22013279 temperature decoding

luoluomeng:
## Summary

Fix AC temperature parsing for model `22013279`, which reports valid temperature data through the B1 `0x7E` message while its generic C0 temperature fields are unreliable.

- Enable the new-protocol `0x7E` temperature path only for model `22013279`, without restricting by subtype.
- Decode indoor temperature from payload bytes `40` and `41`.
- Prevent stale C0 target/indoor/outdoor temperatures from overwriting valid `0x7E` values.
- Add regression coverage for the model-specific gate, subtype-independent handling, and corrected byte offsets.

This PR does not change the AC write/control protocol.

## Background

PR #501 attempted to fix incorrect C0 temperature decoding on the original device by using the B1 `0x7E` response instead. However, it accidentally used `data[39]` and `data[40]` for the indoor-temperature fields.

That incorrect offset caused invalid readings on other devices using the same `0x7E` response. It made the issue appear to be a device-specific `0x7E` format difference.

Based on those symptoms, PR #567 isolated subtype 8 and other unrelated devices. The root cause was instead the incorrect byte offsets introduced in #501. This PR corrects the offsets to `data[40]` and `data[41]`, while limiting the `0x7E` temperature path to the confirmed affected model (`22013279`) so other models retain their existing parsing behavior.

Many thanks to @caibinqing and @wuwentao for identifying this issue. Apologies for the confusion.


wuwentao:
@luoluomeng  thank you very much for your PR and root cause analysis, please don't worry about the bug and issue, new feature always have some bugs, we can't fully cover all the scenarios and all device subtypes.

for this PR review and finally solution, I am still have a new requirements: current solution only apply this feature to a fixed model(prevent unexpected result), if there is any new model have the same features, we will have more delay with new release and code changes, so we can consider to add a customize option to overwrite the default value/result.
customize option should have the first proirity, then check this fixed model value.

the detail solution can be refer to `_refresh_temperature_limits` process in https://github.com/midea-lan/midea-local/blob/feeab8165d796cc1d5b1045f0593724d75d0ed13/midealocal/devices/ac/__init__.py#L545
then we can add this option to `0xAC` device docs, new device and subtype can apply it by a switch. (if you have any good solution, you also can ignore it and do it based on your idea).

luoluomeng:
> @luoluomeng thank you very much for your PR and root cause analysis, please don't worry about the bug and issue, new feature always have some bugs, we can't fully cover all the scenarios and all device subtypes.
> 
> for this PR review and finally solution, I am still have a new requirements: current solution only apply this feature to a fixed model(prevent unexpected result), if there is any new model have the same features, we will have more delay with new release and code changes, so we can consider to add a customize option to overwrite the default value/result. customize option should have the first proirity, then check this fixed model value.
> 
> the detail solution can be refer to `_refresh_temperature_limits` process in
> 
> https://github.com/midea-lan/midea-local/blob/feeab8165d796cc1d5b1045f0593724d75d0ed13/midealocal/devices/ac/__init__.py#L545
> 
> 
> then we can add this option to `0xAC` device docs, new device and subtype can apply it by a switch. (if you have any good solution, you also can ignore it and do it based on your idea).

Thanks for understanding.
Yes, it seems Midea is releasing new devices with different protocols, and the control method still isn't working on this model '22013279'.
A switch would be good at this moment. If I get any new ideas, I'll let you know.

caibinqing (approved):
Thanks for tracking down the real root cause — the off-by-one explains every report in wuwentao/midea_ac_lan#893, including why genuine subtype-8 devices decoded −25.0 (byte 39 happened to be `0x00`).

The model gate is the right call, and my device confirms it's still needed even with the corrected offsets: on 22251759 / subtype 32773, byte 40 is `0x64`, which now decodes to a perfectly plausible 25.0 °C — it would latch and silently drop the C0 outdoor temperature. Gating by verified model closes exactly that hole, and falling back to C0 for the other subtype-8 models (e.g. 00000Q18) matches their logs, where C0 was correct all along.

One minor naming nit, feel free to ignore or defer: the `SUBTYPE8_*` constants and `_parse_subtype8_temperatures` / `has_subtype8_temperature` / `subtype8_temperature` still carry the subtype-8 name, but the gate is now model-based. Renaming to something like `NEW_PROTOCOL_TEMPERATURE_*` would keep the next reader from looking for a subtype check that no longer exists. Fine as a follow-up.

LGTM.

chemelli74:
> One minor naming nit, feel free to ignore or defer: the `SUBTYPE8_*` constants and `_parse_subtype8_temperatures` / `has_subtype8_temperature` / `subtype8_temperature` still carry the subtype-8 name, but the gate is now model-based. Renaming to something like `NEW_PROTOCOL_TEMPERATURE_*` would keep the next reader from looking for a subtype check that no longer exists. Fine as a follow-up.

@caibinqing, please open an issue so this is not forgotten ;-)

PR #590, opened 2026-08-02
test(ac): add coverage for invalid temperature data in subtype-8 0x7E messages

rokam:
## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for invalid temperature data in subtype-8 `0x7E` messages.
  * Verified that no target temperature is exposed when both available decoding methods return invalid data.


chemelli74 (approved):
Thx !

PR #601, opened 2026-08-03
fix(ac): keep unit on when a mode change is followed by a temperature write

### LINKED ISSUE #495: Back-to-back AC mode + target temperature writes can turn unit back off (observed on Midea PortaSplit)
## Summary
When a caller sends two AC writes back-to-back:

1. set HVAC mode to `cool`
2. immediately set target temperature

…the unit can briefly turn on and then immediately turn back off.

I reproduced this reliably with a **Midea PortaSplit** using Home Assistant through `midea_ac_lan`, but the underlying issue appears to be in `midea-local`'s AC write/state handling.

## Environment
- Library: `midea-local==6.8.0`
- Observed through: `midea_ac_lan`
- Device: **Midea PortaSplit**

## Reproduction
Two writes sent with no delay between them:

1. `set_hvac_mode(cool)`
2. `set_temperature(21)`

Equivalent behavior from the HA integration layer:
- `MideaClimate.set_hvac_mode()` -> `self._device.set_attribute(attr="mode", value=...)`
- `MideaClimate.set_temperature()` -> `self._device.set_target_temperature(..., mode=None)`

## Observed behavior
The device state history shows:
- `cool` at `12:22:40.035`
- `off` at `12:22:40.173`

So it flips back to `off` about **138 ms** later.

I also observed the same pattern overnight from an automation:
- AC turned on
- then immediately went back to `off`

## Important control test
If I insert a ~2 second delay between the same two writes:

1. set mode to `cool`
2. wait 2s
3. set target temperature to `21`

…the unit stays on as expected.

So this looks timing-sensitive / state-cache-sensitive, not a general device rejection of the command sequence.

## Suspected root cause
The likely issue is that outbound AC packets are built from cached device attributes that only reflect the **last received state**, not the **latest desired/pending state**.

Relevant code in `midealocal/devices/ac/__init__.py`:

### `set_attribute()`
When changing mode, it builds a full packet and forces `power=True` in the outgoing message:

```python
elif attr in self._attributes:
    message = self.make_message_uniq_set()
    ...
    setattr(message, str(attr), value)
    if attr == DeviceAttributes.mode:
        setattr(message, str(DeviceAttributes.power.value), True)
```

### `set_target_temperature()`
It also builds a full packet from `make_message_uniq_set()`, but only forces `power=True` if `mode is not None`:

```python
message = self.make_message_uniq_set()
message.target_temperature = target_temperature
if mode is not None:
    message.power = True
    message.mode = mode
self.build_send(message)
```

If the first command's response has not yet been processed, the cached attributes may still contain `power=False`. Then the second write can serialize a packet using stale state and effectively turn the unit back off.

The device cache appears to be updated only when inbound responses are parsed in `process_message()` / `update_all()`.

## Why this looks like a library issue
A caller should not need to sleep between two normal sequential writes just to avoid a stale-state race. The library appears to make rapid sequential writes unsafe because it serializes each command from last-confirmed state rather than a pending/desired merged state.

## Proposed fix ideas
Any of these would likely solve it:

1. **Maintain pending/desired outbound state**
   - build packets from last-confirmed state + pending writes
   - clear pending values when confirmed by device response

2. **Optimistically update cached attributes before send**
   - for example, when `set_attribute(attr="mode", ...)` is called, update local cached `mode` and `power=True` before the next packet is built

3. **Provide an atomic AC state setter**
   - e.g. a method that can set mode/power/target temp together in one packet build

## Practical impact
This causes real-world failures in automations that turn on the AC and immediately set the target temperature. In my case it required an HA-side workaround (adding a delay), but it would be much better to fix this in `midea-local`.

If useful, I can also provide a reduced caller-side example, but the timing pattern above was sufficient to reproduce it consistently on the Midea PortaSplit.

wuwentao:
## Summary

Two AC writes sent back-to-back — `set_hvac_mode(cool)` then `set_target_temperature(21)` — could make the unit briefly turn on and then immediately turn **back off** (reported on a Midea PortaSplit; state history showed `cool` → `off` ~138 ms later).

Root cause: `set_attribute("mode", ...)` forces `power=True` **only on the outgoing packet**, not in the cached attributes. The immediately-following `set_target_temperature(mode=None)` builds a *full* packet from `make_message_uniq_set()`, which reads the still-stale last-confirmed `power=False`. That second packet therefore serialized `power=False` and turned the unit off. Inserting a ~2 s delay between the two writes hid the race, because the mode response updated the cache in between.

## Fix

After the mode packet is built (and after the DRY-mode fan reset, which reads the *old* mode), optimistically reflect the commanded state in the cached attributes:

```python
self._attributes[DeviceAttributes.power] = True
self._attributes[DeviceAttributes.mode] = value
```

So an immediate follow-up write is serialized from the intended `power=True` + new mode instead of the stale confirmed state. The real device response still overwrites the cache when it arrives (matching the optimistic-update pattern the integration already relies on).

## Changes

- `midealocal/devices/ac/__init__.py`: optimistic cache update in the `mode` branch of `set_attribute`.

## Tests

- Added `test_set_mode_then_temperature_keeps_unit_on`: from a confirmed `power=False, mode=0` state, `set_attribute("mode", cool)` then `set_target_temperature(21, None)` — asserts both packets carry `power=True`, mode `cool`, and the cache is updated. Verified this test **fails without the fix** (the follow-up packet's `power` was `False`) and passes with it.

## Verification

- `uv run python -m pytest tests/devices/ac/` — 131 passed
- `uvx ruff@0.6.9 check` / `format --check` — clean
- `uv run mypy midealocal` — Success
- `uv run pylint --rcfile=pylintrc midealocal` — 10.00/10

Fixes #495

## Summary by CodeRabbit

* **Bug Fixes**
  * Changing the air conditioner’s mode now immediately reflects the selected mode and power-on state.
  * Setting the temperature immediately after changing modes now preserves the latest commands instead of using stale device state.

* **Tests**
  * Added regression coverage for consecutive mode and temperature changes.

### DIFF
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -857,6 +857,15 @@ class MideaACDevice(MideaDevice):
                     # Force fan_speed to AUTO when leaving DRY mode
                     if self._attributes[DeviceAttributes.mode] == DRY_MODE:
                         message.fan_speed = 102
+                    # Optimistically reflect the commanded state in the cache so
+                    # an immediate follow-up write (e.g. set_target_temperature,
+                    # which serializes a full packet from make_message_uniq_set)
+                    # is built from power=True and the new mode. Without this the
+                    # follow-up reuses the stale last-confirmed power=False and
+                    # turns the unit back off before the mode response arrives.
+                    # https://github.com/midea-lan/midea-local/issues/495
+                    self._attributes[DeviceAttributes.power] = True
+                    self._attributes[DeviceAttributes.mode] = value
         if message is not None:
             self.build_send(message)
 
--- a/tests/devices/ac/device_ac_test.py
+++ b/tests/devices/ac/device_ac_test.py
@@ -230,6 +230,36 @@ class TestMideaACDevice:
             assert message.dry is False
             assert message.fan_speed == 102
 
+    def test_set_mode_then_temperature_keeps_unit_on(self) -> None:
+        """A mode change followed immediately by a temperature write stays on.
+
+        set_attribute("mode") only forced power=True on the outgoing packet, so
+        a rapid follow-up set_target_temperature (built from make_message_uniq_set)
+        reused the stale last-confirmed power=False and turned the unit back off.
+        The mode change now optimistically caches power=True + the new mode.
+        https://github.com/midea-lan/midea-local/issues/495
+        """
+        # Start from the confirmed "off" state, as after a fresh query.
+        self.device._attributes[DeviceAttributes.power] = False
+        self.device._attributes[DeviceAttributes.mode] = 0
+        with patch.object(self.device, "build_send") as mock_build_send:
+            # 1. set HVAC mode to cool (value 2)
+            self.device.set_attribute(DeviceAttributes.mode.value, 2)
+            mode_message = mock_build_send.call_args[0][0]
+            assert mode_message.power is True
+            assert mode_message.mode == 2
+            # Cache reflects the commanded state before the device responds.
+            assert self.device.attributes[DeviceAttributes.power] is True
+            assert self.device.attributes[DeviceAttributes.mode] == 2
+
+            # 2. immediately set target temperature (mode=None, as the climate
+            #    entity does) - must not resurrect the stale power=False.
+            self.device.set_target_temperature(21, None)
+            temp_message = mock_build_send.call_args[0][0]
+            assert temp_message.power is True
+            assert temp_message.mode == 2
+            assert temp_message.target_temperature == 21
+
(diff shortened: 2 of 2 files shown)


rokam (commented):
I think this will do the job, although if the send packet action fails. The state of the device will be wrong. 

Eg. The device is off and the mode is set to Cool. But the packet is lost due to LAN connection and UDP protocol. The device will appear to be On and Cool even when it is really off. This will remains until next pull.

I think the solution needs to be more robust.

chemelli74 (dismissed):
Please address open comments

rokam (approved):
Thinking better, I do belive this is the best solution. It's better to avoid this then the problem I did point.

PR #574, opened 2026-07-30
chore(main): release midea-local 7.0.0

wuwentao:
:robot: I have created a release *beep* *boop*
---

### ⚠ BREAKING CHANGES

* **b8:** drop const and align behavior with other devices ([#579](https://github.com/midea-lan/midea-local/issues/579))
* **c3:** drop const in devices ([#577](https://github.com/midea-lan/midea-local/issues/577))

### Features

* **b0:** map mode 0xA1 time defrost in _mode31 ([#570](https://github.com/midea-lan/midea-local/issues/570)) ([514a583](https://github.com/midea-lan/midea-local/commit/514a583c1f8cab1dcf8fb926ac7779234b186607))

### Bug Fixes

* **ac:** expose named speeds for custom fan capability ([#568](https://github.com/midea-lan/midea-local/issues/568)) ([fb5e7d7](https://github.com/midea-lan/midea-local/commit/fb5e7d7b2f1bf16b8ca7e69101722b033a69390b))
* **ac:** make screen_display switch idempotent ([#600](https://github.com/midea-lan/midea-local/issues/600)) ([05e5f95](https://github.com/midea-lan/midea-local/commit/05e5f95f562bd71c57be1f62807320114942677c))
* **ac:** model 22013279 temperature decoding ([#572](https://github.com/midea-lan/midea-local/issues/572)) ([efae776](https://github.com/midea-lan/midea-local/commit/efae776232a2baacd41f7284c17b8ebb30022a57))
* **b0:** assemble cloudmenuid as a big-endian 3-byte value ([#569](https://github.com/midea-lan/midea-local/issues/569)) ([cb3ec5e](https://github.com/midea-lan/midea-local/commit/cb3ec5ecfeecf1dc68926fa03e0de567ba9526b6))

### Code Refactoring

* **b8:** drop const and align behavior with other devices ([#579](https://github.com/midea-lan/midea-local/issues/579)) ([f233f15](https://github.com/midea-lan/midea-local/commit/f233f15c246328275e26b1d52efaf948ccd8e132))
* **c3:** drop const in devices ([#577](https://github.com/midea-lan/midea-local/issues/577)) ([752ef52](https://github.com/midea-lan/midea-local/commit/752ef52e32d40311eb356b15e3bcaa22a88cfc0c))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #605, opened 2026-08-04
refactor(ac)!: rename subtype8 temperature names to new_protocol

caibinqing:
## Summary

Follow-up to #572, as requested in https://github.com/midea-lan/midea-local/pull/572#issuecomment-5171878028.

The 0x7e temperature decoding gate has been model-based since #572, but the constants and helpers in `message.py` still carried the subtype-8 name. This renames them to align with the existing device-side naming (`NEW_PROTOCOL_TEMPERATURE_MODELS`, `_uses_new_protocol_temperature`), so the next reader does not look for a subtype check that no longer exists.

Pure rename, no behavior change:

- `SUBTYPE8_*` constants -> `NEW_PROTOCOL_*`
- `_parse_subtype8_temperatures` -> `_parse_new_protocol_temperatures`
- `has_subtype8_temperature` -> `has_new_protocol_temperature`
- `subtype8_temperature` kwarg -> `new_protocol_temperature`

Also updated one test docstring that referenced the old name.

## Test

- `pytest tests/devices/ac`: 131 passed
- ruff check / ruff format / mypy / pylint (pre-commit): all pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved air-conditioner temperature reading selection by prioritizing newer-protocol temperature data over stale readings.
  * Preserved accurate setpoint and indoor-temperature decoding, validation, fallback behavior, and outdoor-temperature handling.

* **Refactor**
  * Updated temperature terminology to reflect the newer protocol more accurately.
  * No expected change to existing user-facing temperature behavior beyond improved reading selection.


chemelli74:
Thx for addressing the comment.

I think this should be marked as a breaking change.

Correct ?

caibinqing:
Strictly by semver, yes — these names shipped in v6.11.1, so the rename is technically breaking for anyone importing them directly. In practice nobody does: the known consumers go through `MideaACDevice` and pin exact versions. Note #572 already renamed the device-side gate and removed the released `SUBTYPE8` constant without a `!`, so whatever we decide should apply to both.

Since #574 is already at 7.0.0, marking this is essentially free if it lands before that release is cut. No strong opinion — feel free to edit the title.

chemelli74:
Unfortunately, I think there will be a few more breaking changes over the coming days as we continue to make the codebase clearer and more readable.

So I think we should mark this as a breaking change as well.

**Note:** Sorry, I saw your message only after releasing 7.0.0. I was focused on fixing `release-please`. 😞

chemelli74:
@caibinqing O marked a couple changes as breaking.
I think you can proceed with this one as well.

Thank you !

caibinqing:
Done — retitled with `!`.

PR #618, opened 2026-08-06
refactor(device): share attribute update implementation across devices

chemelli74:
- Added MideaDevice.update_attributes_from_message() in midealocal/device.py — a shared helper that replaces the "for each declared attribute, copy it from the response and record it in both self._attributes and new_status" loop that was hand-written 30 times.
- Migrated 30 devices to call it: ad, b0, b1, b3, b4, b8, bf, c2, c3, ca, ce, cf, da, db, dc, e1, e2, e3, e6, e8, ea, ec, ed, fb, fc, fd, x13, x26, x34, x40.
- Left 6 devices untouched, each for a documented reason (in the plan file): ac, cc, cd, b6 (cross-attribute/order-dependent logic), a1, fa (a small side-effect that would need extra care to replicate exactly).
- No test changes were required, confirming that the refactor only changes code structure without affecting behavior.

Net effect: -206 lines across 31 files, zero behavior change otherwise.

EDIT:

Added also the last 6 devices to the list ! Now refactor is complete 🔝 

## New Features
- Improved device status processing across supported appliances.
- Added more consistent conversion of modes, fan speeds, temperatures, progress, and other reported values.
- Improved handling of device-specific status mappings and fallback values.

## Bug Fixes
- Improved synchronization of reported device states.
- Ignored stale or invalid readings without overwriting valid state.
- Preserved cached values when status messages omit fields.
- Correctly reset fan speed when devices report powering off.
- Control responses are returned promptly where supported.


chemelli74:
@wuwentao, @caibinqing, and @rokam, could you please give this one a quick test before I merge it?

I'm 99% confident that everything will work as expected, but having a confirmation from you would make me even more confident.

Thanks in advance!

midealocal/devices/da/__init__.py
```diff
@@ -121,39 +120,30 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+                DeviceAttributes.detergent: lambda v: (
+                    None if v >= len(detergent) else detergent[v]
+                ),
+                DeviceAttributes.softener: lambda v: (
+                    None if v >= len(softener) else softener[v]
+                ),
+                DeviceAttributes.wash_strength: lambda v: (
+                    None if v >= len(strength) else strength[v]
+                ),
+            },
```
rokam: I don't like those lamba functions. It's hard to read. Maybe create another helper to parse values from dict and arrays.

midealocal/devices/e3/__init__.py
```diff
@@ -82,19 +82,16 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
                     DeviceAttributes.target_temperature,
-                ]:
-                    self._attributes[status] = getattr(message, str(status)) / 2
-                else:
-                    self._attributes[status] = getattr(message, str(status))
-                new_status[str(status)] = self._attributes[status]
-
-        return new_status
+                ],
+                lambda v: v / 2 if self._precision_halves else v,
```
rokam: A helper for precision_halves can also be usefull

midealocal/device.py
```diff
@@ -631,6 +636,37 @@ def update_all(self, status: dict[str, Any]) -> None:
         for update in self._updates:
             update(status)
 
+    def update_attributes_from_message(
+        self,
+        message: object,
+        translators: dict[str, Callable[[Any], Any]] | None = None,
+        default_transform: Callable[[Any], Any] | None = None,
+    ) -> dict[str, Any]:
```
rokam, line 734: I think we could use better typing here for those callable. They should return  [str|int|float|bool] .

rokam (dismissed):
I do think it is hard to read using those lambda functions. To reduce code dup we could use more helpers instead of lamda.

midealocal/devices/da/__init__.py
chemelli74: Agree, needs a bit of rework before merging.

Current code is working for you ?
So at least we know the starting point is a good one

rokam: I haven't tested yet

midealocal/device.py
chemelli74: True, done.

Note: `float` already include `int`.

midealocal/devices/e3/__init__.py
chemelli74: Done

midealocal/devices/da/__init__.py
chemelli74: Lambdas removed 😃

midealocal/devices/fa/__init__.py
```diff
@@ -139,47 +139,42 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+                    if v < len(MideaFADevice._oscillation_modes)
+                    else None
+                ),
+                DeviceAttributes.mode: lambda v: (
+                    MideaFADevice._modes[v] if v < len(MideaFADevice._modes) else None
+                ),
+                DeviceAttributes.power: _translate_power,
+                DeviceAttributes.fan_speed: _translate_fan_speed,
+            },
+        )
```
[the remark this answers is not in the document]
chemelli74: @coderabbitai, check after last commit

midealocal/devices/fd/__init__.py
```diff
@@ -116,31 +116,18 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
-                new_status[str(status)] = self._attributes[status]
-        return new_status
+        return self.update_attributes_from_message(
+            message,
+            {
+                DeviceAttributes.mode: lambda v: (
+                    MideaFDDevice._modes[v - 1]
+                    if v <= len(MideaFDDevice._modes)
+                    else None
+                ),
```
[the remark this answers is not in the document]
chemelli74: @coderabbitai, check after last commit

midealocal/devices/fa/__init__.py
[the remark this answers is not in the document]
chemelli74: @coderabbitai, check after last commit

chemelli74:
@coderabbitai, full review

chemelli74:
@wuwentao, may I kindly request a feedback ?

midealocal/device.py
```diff
@@ -39,6 +39,100 @@
...
+
+
+def list_translator(
+    values: Sequence[str],
+    *,
+    offset: int = 0,
+    min_index: int = 0,
+    default: str | None = None,
+    key: Callable[[int], int] | None = None,
+) -> Callable[[int], str | None]:
```
rokam, line 56: Please create unit tests for this method. That covers each type of args usage.

midealocal/device.py
```diff
@@ -39,6 +39,100 @@
...
+        i = (key(index) if key is not None else index) - offset
+        return values[i] if min_index <= i < len(values) else default
+
+    return _translate
+
+
+def dict_translator(
+    mapping: Mapping[Any, Any],
+    default: Any = _NO_DEFAULT,  # noqa: ANN401
+) -> Callable[[Any], Any]:
```
rokam, line 79: Please create unit tests for this method. That covers each type of args usage.

midealocal/device.py
```diff
@@ -39,6 +39,100 @@
...
+
+    def _translate(value: Any) -> Any:  # noqa: ANN401
+        return mapping.get(value, value if default is _NO_DEFAULT else default)
+
+    return _translate
+
+
+def precision_halves_translator(
+    precision_halves: bool | None,
+) -> Callable[[float], float]:
```
rokam, line 94: Please create unit tests for this method. That covers each type of args usage.

midealocal/device.py
```diff
@@ -39,6 +39,100 @@
...
+    value through unchanged.
+    """
+
+    def _translate(value: float) -> float:
+        return value / 2 if precision_halves else value
+
+    return _translate
+
+
+def multiplier_translator(multiplier: float) -> Callable[[float | None], float | None]:
```
rokam, line 109: Please create unit tests for this method. That covers each type of args usage.

midealocal/device.py
```diff
@@ -39,6 +39,100 @@
...
+
+    def _translate(value: float | None) -> float | None:
+        if value is None or multiplier == 1.0:
+            return value
+        return round(value * multiplier)
+
+    return _translate
+
+
+def sentinel_translator(sentinel: Any, replacement: Any) -> Callable[[Any], Any]:  # noqa: ANN401
```
rokam, line 124: Please create unit tests for this method. That covers each type of args usage.

rokam (dismissed):
I think some tests should guarantee that thos transformers should work as expected.
Although the lines are covered by device tests I think we need better unit tests for these main methods.

All the rest looks good for me.

rokam (dismissed):
LGTM

rokam (approved):

PR #619, opened 2026-08-06
fix(ac): report self-clean status correcly

### LINKED ISSUE #971: self_clean permanently reports "on" on PortaSplit (subtype 524) — byte-12 decoding from midea-local PR #467 does not hold
### Verify Issue Submission

- [x] I have read the [README documentation](https://github.com/wuwentao/midea_ac_lan/blob/main/README.md) , understand how to use midea_ac_lan and enable debug log.
- [x] I have read the [device documentation](https://github.com/wuwentao/midea_ac_lan/blob/main/README.md#4-supported-appliances) for my device in the README, but the problem still can't fix.
- [x] I have submitted the device type and SN information.
- [x] I have uploaded the full debug log file, rather than just paste a parts of log based on my own judgment.
- [x] I have accurately and clearly described what the current error is and what the expected correct result should be.
- [x] I have searched for my issue in the [Issue Tracker](https://github.com/wuwentao/midea_ac_lan/issues) and have not found it.

### HA core version

2025.12.4

0.7.1

unknown

### Device type and Device SN

000000P0000000Q1D450EEFA7AFA0000

### Working mobile app

SmartHome

### The detail description of bug/problem

> **Note on the attached log:** I am attaching a filtered debug log containing only the `midea` entries (2071 of 4016 lines, covering 09:52–10:27). The full log also contains unrelated integrations, internal IP addresses and automation names from my home, which I would rather not publish. I verified the filtered file contains no token/key material. If you need the complete file, I can send it privately.

**1. Device type and name**

Two Midea PortaSplit portable split units, on two separate Home Assistant instances:

| Unit | Model | Subtype | Protocol |
|---|---|---|---|
| A (log attached) | `00000Q1D` | 524 | 3 |
| B | `2060008E` | 524 | 3 |

Integration `midea_ac_lan` v0.7.1, protocol layer `midea-local` 6.11.1 (as pinned in `manifest.json`).

**2. Error action / error info**

`switch.<id>_self_clean` reports `on` permanently on **both** units, while no self-cleaning cycle is running.

Verified at the appliance, not only in the UI:

- The indoor unit shows **no `CL`** on its display.
- The Midea SmartHome app shows self-clean as **off**.
- Both units report `on` continuously, across restarts, with `power` on and the unit cooling normally.

Turning the switch off in Home Assistant therefore sends an **abort command for a cycle that is not running**, and the entity cannot be used as a status indicator or as an automation condition. Starting a cycle still works — only the reported state is wrong.

**3. Summary, expected result, and why I believe the decoding is at fault**

Expected: `self_clean` reflects whether a self-cleaning cycle is actually running — `off` while the unit is cooling and the display shows no `CL`.

Command and status use two different tags:

- **Command** — tag `0x0039`, payload `0x01` start / `0x00` abort. The source itself notes this channel cannot be read back: `# self_clean query can't return response`
- **Status** — tag `0x00E2` (`b5_self_clean_active`), decoded in `midealocal/devices/ac/message.py` as `len(data) > 12 and data[12] != 0`

The attribute is assigned in `midealocal/devices/ac/__init__.py`:

```python
if hasattr(message, "self_clean_active"):
    active = message.self_clean_active
    self._attributes[DeviceAttributes.self_clean] = active
```

`set_attribute` does not optimistically set a local value for `self_clean`, so the `on` state can only originate from a `0x00E2` frame with byte 12 != 0. It is not an echo of a command I sent.

That decoding was introduced in https://github.com/midea-lan/midea-local/pull/467, described there as derived "after further investigation with a live device" — a single unit, with no format validation of the 39-byte payload. `mill1000/midea-msmart` obtains the same information by reading tag `0x0039` itself and does not use `0x00E2` at all. Two independent implementations disagreeing about the status channel suggests the rule is not general.

On this firmware, byte 12 appears to carry something other than a self-clean flag. The attached debug log contains the raw frames from unit A.

**Suggested direction:** validate the `0x00E2` payload before trusting byte 12, gate the status decoding behind a capability or firmware check, or follow msmart and read the state back from `0x0039`. Splitting the entity into a `button` (start) plus a `binary_sensor` (running) would also match the fact that self-cleaning is a one-shot program, not a held state.

I can capture further logs from either unit on request.

### The logs

[midea-debug-bereinigt.log](https://github.com/user-attachments/files/30736256/midea-debug-bereinigt.log)

tilalx:
## Summary

Fixes incorrect Self Clean status reporting for PortaSplit subtype 524.

`0x00E2` is a 39-byte status payload. The previous decoder treated byte 12 as
a boolean, but live captures show it remains non-zero when no self-clean cycle
is running. The actual running flag is byte 13, bit `0x40`.

This also change handles delayed status responses after a Self Clean command:

- Publish the requested switch state after the command is sent successfully.
- Ignore contradictory `0x00E2` reports until the device confirms the requested
  state or one configured refresh interval expires.
- Once confirmed, accept later status changes normally, including cycle completion.

## Validation

- Live-tested against a PortaSplit 

Fixes wuwentao/midea_ac_lan#971

## Summary by CodeRabbit

* **New Features**
  * Added dedicated self-clean status querying for supported air conditioners.
  * Self-clean commands now provide immediate state updates while awaiting confirmation.

* **Bug Fixes**
  * Prevented outdated self-clean responses from overwriting current pending states.
  * Added timeout handling to refresh self-clean status when confirmation is delayed.
  * Failed self-clean commands no longer incorrectly change the reported state.

* **Tests**
  * Expanded coverage for self-clean queries, command success and failure, delayed responses, and capability reporting.

### DIFF
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -512,7 +512,7 @@ class MideaACDevice(MideaDevice):
             if self._pending_self_clean is not None:
                 expected, set_at = self._pending_self_clean
                 elapsed = time.monotonic() - set_at
-                if active == expected or elapsed > self._self_clean_pending_timeout:
+                if active == expected or elapsed >= self._self_clean_pending_timeout:
                     self._pending_self_clean = None
                 else:
                     _LOGGER.debug(
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+import time
 from dataclasses import dataclass
 from enum import StrEnum
 from typing import Any, ClassVar, Unpack, cast
@@ -311,6 +312,7 @@ class MideaACDevice(MideaDevice):
             dict.fromkeys(self._model_capabilities.attributes),
         )
         self._fresh_air_version: DeviceAttributes | None = None
+        self._pending_self_clean: tuple[bool, float] | None = None
         self._default_temperature_step: float = 0.5
         self._temperature_step: float = 0.5
         self._used_subprotocol: bool = self._model_capabilities.uses_bb_protocol
@@ -502,8 +504,26 @@ class MideaACDevice(MideaDevice):
             self._fresh_air_version = DeviceAttributes.fresh_air_2
         if hasattr(message, "self_clean_active"):
             active = message.self_clean_active
-            self._attributes[DeviceAttributes.self_clean] = active
-            new_status[DeviceAttributes.self_clean.value] = active
+            update_self_clean = True
+            if self._pending_self_clean is not None:
+                expected, set_at = self._pending_self_clean
+                now = time.monotonic()
+                elapsed = now - set_at
+                if active == expected:
+                    self._pending_self_clean = None
+                elif elapsed <= self._self_clean_pending_timeout:
+                    _LOGGER.debug(
+                        "[%s] Ignoring stale self-clean status %s while awaiting %s",
+                        self.device_id,
+                        active,
+                        expected,
+                    )
+                    update_self_clean = False
+                else:
+                    self._pending_self_clean = None
+            if update_self_clean:
+                self._attributes[DeviceAttributes.self_clean] = active
+                new_status[DeviceAttributes.self_clean.value] = active
         new_status.update(self._refresh_temperature_limits(message))
         self._update_capabilities(message)
         return new_status
@@ -523,6 +543,15 @@ class MideaACDevice(MideaDevice):
                 break
             mode = name
         return mode
+
+    @property
+    def _self_clean_pending_timeout(self) -> int:
+        """Return the stale-status window for a pending self-clean command."""
+        return (
+            self._refresh_interval
+            if self._refresh_interval > 0
+            else self._default_refresh_interval
+        )
 
     def _update_capabilities(self, message: MessageACResponse) -> None:
         """Accumulate decoded B5 capability flags from a B5 response."""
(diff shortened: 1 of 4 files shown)


midealocal/devices/ac/__init__.py
```diff
@@ -502,8 +504,26 @@ def process_message(self, msg: bytes) -> dict[str, Any]:  # noqa: C901
...
+                elif elapsed <= self._self_clean_pending_timeout:
+                    _LOGGER.debug(
+                        "[%s] Ignoring stale self-clean status %s while awaiting %s",
+                        self.device_id,
+                        active,
+                        expected,
+                    )
+                    update_self_clean = False
+                else:
+                    self._pending_self_clean = None
```
chemelli74: You can make it simpleer and cleaner:

```
if active == expected or elapsed > self._self_clean_pending_timeout:
   self._pending_self_clean = None
else:
  ....
```

I think also that the best practice is to use datetime instead of time, needs to check the difference

chemelli74 (dismissed):

midealocal/devices/ac/__init__.py
tilalx: I checked, and I'd like to keep  time.monotonic()  here. The stdlib [docs](https://docs.python.org/3/library/time.html#time.monotonic) are actually explicit about this being the intended use case

midealocal/devices/ac/__init__.py
```diff
@@ -502,8 +508,23 @@ def process_message(self, msg: bytes) -> dict[str, Any]:  # noqa: C901
...
         if hasattr(message, "self_clean_active"):
             active = message.self_clean_active
-            self._attributes[DeviceAttributes.self_clean] = active
-            new_status[DeviceAttributes.self_clean.value] = active
+            update_self_clean = True
+            if self._pending_self_clean is not None:
+                expected, set_at = self._pending_self_clean
+                elapsed = time.monotonic() - set_at
+                if active == expected or elapsed > self._self_clean_pending_timeout:
+                    self._pending_self_clean = None
```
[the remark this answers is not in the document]
chemelli74: This was marked resolved with no code change or explanation.
Why ?

PR #621, opened 2026-08-06
fix(e2): encode subtype 255 temperature literally

yeaxi:
Hello!

I noticed an issue with my Midea boiler where I couldn't set the target temperature. It would always try to set it to max temp and then reset back to previous value. I've found the cause and prepared a fix for it.

I have a Midea E2 devices with `subtype=255`.

These devices expect the temperature as a literal Celsius value, while the existing default path multiplied it by `2` for the legacy half-degree representation. As a result, commands such as `45°C` were sent as `90`, rejected by the device, and reverted on the next state update.

For `subtype=255`, the fix now sends the literal value while preserving the existing behavior for other E2 variants and for `precision_halves`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected target-temperature handling for E2 subtype 255 devices.
  * Subtype 255 devices now receive literal Celsius values, including when half-degree precision is disabled.

* **Tests**
  * Added regression coverage to verify the expected serialized temperature command.

### DIFF
--- a/midealocal/devices/e2/__init__.py
+++ b/midealocal/devices/e2/__init__.py
@@ -83,6 +83,12 @@ class DeviceAttributes(StrEnum):
     rate = "rate"
 
 
+# E2 subtype 255 devices use literal Celsius values in the target-temperature
+# command.  Other E2 variants use the legacy half-degree wire representation
+# unless ``precision_halves`` is explicitly enabled.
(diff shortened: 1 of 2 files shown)


rokam (approved):
LGTM

chemelli74 (approved):
LGTM, thx !

PR #627, opened 2026-08-08
fix(cli): namespace guard to get_sn attr

rokam:
When running cli with setattr, discover was called without get_sn attr inside namespace

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the `discover` command’s handling of optional configuration values, preventing errors when the value is unavailable.


chemelli74 (approved):
Another nice catch !

PR #629, opened 2026-08-09
fix(b0): ignore 31 body on subtype zero devices

matvejalekseev:
## What
  
  On subtype-zero B0 devices, the 31-body response (`B0Message31Body`) was being
  decoded and applied on top of whatever was just received from the 01-body
  response, even though its byte layout is specific to model 0TG025JG,
  subtype 2 (see the existing comments in `process_message`). On a subtype-zero
  device this produces meaningless values for `door`, `status`, 
  `tank_ejected`, `water_shortage`, `water_change_reminder` and
  `time_remaining` that immediately overwrite the good data the 01-body
  response provided a moment earlier - on every single polling cycle.
  
  ## Why
  
  Confirmed against a real subtype-zero microwave (`model 70000209`) via debug
  logs: every ~10s poll sends both `MessageQuery01` and `MessageQuery31`, and
  the 31-body response - decoded with the wrong (subtype-2) layout - clobbers
  the correct 01-body data essentially all the time. In practice this made
  `door`/`status`/etc. show stale/wrong values almost continuously, with only
  a ~100ms window of correct data right after the 01 response before the 31
  response overwrote it.
  
  ## Fix
  
  Skip processing the 31 body entirely when `subtype == 0`, since the 01 body
  already reliably provides everything this device class reports. `mode`/
  `fire_power`/`child_lock` (only ever present in the 31 body) simply stay
  unset on subtype-zero devices, same as before this change on any occasion
  the 31 body wasn't already corrupting other fields.
  
  ## Testing
  
  - Added `test_process_message_31_ignored_on_subtype_zero`, reproducing the
    scenario: a valid 01-body update followed by a 31-body response that would
    previously have overwritten it.
  - Full suite passes (`pytest ./tests/`).
  - `ruff check`, `ruff format --check`, `mypy`, `pylint` all clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved status handling for subtype-zero devices by ignoring unsupported responses.
  * Preserved valid device attributes received from supported responses.
  * Added regression coverage to ensure device status remains accurate.

### DIFF
--- a/midealocal/devices/b0/__init__.py
+++ b/midealocal/devices/b0/__init__.py
@@ -6,6 +6,7 @@ from typing import ClassVar, Unpack
 
 from midealocal.const import DeviceType
 from midealocal.device import MideaDevice, MideaDeviceInitKwargs
+from midealocal.message import ListTypes
 
 from .message import (
     MessageB0Response,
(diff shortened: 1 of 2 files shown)


wuwentao:
thanks for your PR.
I have 0TG025JG device, so I just add these feature and code comments.
we can merge it and if there is any new issue, we can continue improve it.

PR #490, opened 2026-07-02
chore(main): release 6.11.0

wuwentao:
:robot: I have created a release *beep* *boop*
---

### Features

* **ac:** add group data ([#507](https://github.com/midea-lan/midea-local/issues/507)) ([c8a5a27](https://github.com/midea-lan/midea-local/commit/c8a5a2735a7284ac9eeab7d4e102a8cc1fc03000))
* **ac:** add model-specific airflow and diagnostics ([#502](https://github.com/midea-lan/midea-local/issues/502)) ([c8d1e33](https://github.com/midea-lan/midea-local/commit/c8d1e33eeb54c3d24caeb8ba9caad735776b52d9))
* **ac:** add power saving mode ([#500](https://github.com/midea-lan/midea-local/issues/500)) ([c40b39a](https://github.com/midea-lan/midea-local/commit/c40b39acacd2e8ab129126c6818111ae8f14f81a))
* **ac:** add rate_select (power rate limit / Gen mode) support ([#469](https://github.com/midea-lan/midea-local/issues/469)) ([3e4322d](https://github.com/midea-lan/midea-local/commit/3e4322dfd49ff9cc870e088b4b7a73b467367fdf))
* **e1:** add dishwasher work controls ([#475](https://github.com/midea-lan/midea-local/issues/475)) ([d46f857](https://github.com/midea-lan/midea-local/commit/d46f857722c5b1588fa7ff56545b40d84289b93b))
* **ed:** add soft water machine (subtype 703) support ([#505](https://github.com/midea-lan/midea-local/issues/505)) ([9c11612](https://github.com/midea-lan/midea-local/commit/9c11612a6a13f88c98b560e6ebf3254150e9919f))
* **fa:** add humidify, waterions, display_on_off (AAF10MR, subtype 0) ([#466](https://github.com/midea-lan/midea-local/issues/466)) ([80d67f6](https://github.com/midea-lan/midea-local/commit/80d67f67bc51626f93ce30223263caf04419b072))
* report MAC address ([#488](https://github.com/midea-lan/midea-local/issues/488)) ([cd862d3](https://github.com/midea-lan/midea-local/commit/cd862d39df0dee3a8cb32f9a8d1ec182d86a2840))
* report serial number ([#493](https://github.com/midea-lan/midea-local/issues/493)) ([bef1082](https://github.com/midea-lan/midea-local/commit/bef1082822a5d51b5e5924fc1f26fa089f5ff972))

### Bug Fixes

[...]

PR #630, opened 2026-08-09
refactor(ac)!: rename T1/T2/T3 group 1 temperature attributes

s00500:
Follow-up to the group 1/2/7 diagnostics added in #490, from review feedback by
@TurboLed on wuwentao/midea_ac_lan#887.

The three refrigerant-circuit temperatures were named after their role in cooling
mode, which is wrong half the time:

| | before | after | why |
| --- | --- | --- | --- |
| T1 | `indoor_coil_temperature` | `indoor_ambient_temperature` | T1 is the return-air thermistor - the air around the indoor unit, not the coil |
| T2 | `evaporator_temperature` | `indoor_coil_temperature` | the indoor coil is the evaporator only while cooling; in heat mode it condenses |
| T3 | `condenser_temperature` | `outdoor_coil_temperature` | same, inverted - the outdoor coil is the condenser only while cooling |

Naming by location is mode-independent and matches the service-manual convention
(see [SG-RG10-01](https://cematraining.com/wp-content/uploads/2022/03/SG-RG10-01.pdf)).
T4 `outdoor_ambient_temperature` and TP `discharge_pipe_temperature` were already
location-named and are untouched. The CD (heat-pump water heater) attribute
`condenser_temperature` is a different device and is untouched.

## Summary by CodeRabbit

* **Updates**
  * Renamed air-conditioner temperature readings for clearer sensor identification.
  * Indoor readings are now reported as ambient and coil temperatures.
  * Outdoor readings are now reported as outdoor coil temperature.
  * Updated device controls and read-only sensor handling to use the revised names.
* **Bug Fixes**
  * Improved consistency between reported temperature data and device sensor labels.

### DIFF
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -113,9 +113,9 @@ class DeviceAttributes(StrEnum):
     target_compressor_frequency = "target_compressor_frequency"
     compressor_current = "compressor_current"
     compressor_voltage = "compressor_voltage"
-    indoor_coil_temperature = "indoor_coil_temperature"  # T1
-    evaporator_temperature = "evaporator_temperature"  # T2
-    condenser_temperature = "condenser_temperature"  # T3
+    indoor_ambient_temperature = "indoor_ambient_temperature"  # T1
+    indoor_coil_temperature = "indoor_coil_temperature"  # T2
+    outdoor_coil_temperature = "outdoor_coil_temperature"  # T3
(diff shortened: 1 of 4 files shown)


s00500:
Awesome! Thanks!

Taudris:
Good change to have overall, but I wish it included logic to rename the existing entities so that their data doesn't get lost. Renaming and deleting by hand is painful.

PR #632, opened 2026-08-11
feat(ac): gate rate_select query behind b5_electricity capability

vaughnitem:
## Summary

Follow-up to #469. The B1 (`MessageNewProtocolQuery`) query always requested `rate_select` (tag `0x0048`) regardless of whether the device actually supports it. Devices that don't advertise the capability never answer that tag, so every poll cycle wasted a query slot on unsupported units.

This adds a logic gate: `rate_select` is only queried once the device has advertised support via the B5 `b5_electricity` capability (tag `0x0216`).

### Changes

**`midealocal/devices/ac/message.py`**
- Add `B5_ELECTRICITY_UNSUPPORTED_VALUE = 0` constant.
- `XB5MessageBody._parse_capabilities`: decode `b5_electricity` into `capabilities["rate_select"]` — a nonzero raw value (rate level count) means supported, `0` means unsupported.
- `MessageNewProtocolQuery.__init__` now accepts a `supports_rate_select: bool = False` keyword-only flag; the `rate_select` tag is only appended to the B1 query param list when it's `True`.

**`midealocal/devices/ac/__init__.py`**
- `build_query()` passes `supports_rate_select=self._capabilities.get("rate_select", False)` into `MessageNewProtocolQuery`, so the query reflects whatever the device has reported via B5 capability frames so far (defaults to `False`/excluded before any B5 response is seen).

### Testing

- `tests/devices/ac/message_ac_test.py`:
  - Updated `test_new_protocol_query_body` — default query now excludes `rate_select`.
  - Added `test_new_protocol_query_body_includes_rate_select_when_supported` — verifies the tag is appended when the flag is set.
  - Added `test_message_query_b5_electricity_gates_rate_select` (parametrized) — verifies raw `b5_electricity` values of `4`, `1`, and `0` map to `capabilities["rate_select"]` of `True`, `True`, and `False`.
- `tests/devices/ac/device_ac_test.py`:
  - Added `test_build_query_omits_rate_select_until_capability_confirmed` — verifies `build_query()` excludes `rate_select` from the B1 body when `_capabilities` is empty, and includes it once `_capabilities["rate_select"] = True`.

Ran the full suite plus lint/type-check on a Python 3.12 `uv` venv:
- `uv run python -m pytest ./tests/` → **1478 passed**
- `uv run ruff check .` → clean
- `uv run ruff format --check .` → clean
- `uv run mypy midealocal/devices/ac/message.py midealocal/devices/ac/__init__.py` → no issues
- `uv run pylint --rcfile=pylintrc midealocal/devices/ac/message.py midealocal/devices/ac/__init__.py` → 10.00/10

## Bug Fixes

* Improved air conditioner capability detection for rate selection.
* Rate selection requests are now sent only when the device reports support.
* Correctly handles devices with unavailable or unsupported electricity values.
* Prevents unsupported rate selection options from appearing or being requested.
* Ensures compatible devices receive accurate rate selection information without affecting devices that do not advertise the capability.
* Improved compatibility with devices using different electricity capability values.

### DIFF
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -381,7 +381,10 @@ class MideaACDevice(MideaDevice):
             ]
         queries: list[ACQuery] = [
             MessageQuery(self._message_protocol_version),
-            MessageNewProtocolQuery(self._message_protocol_version),
+            MessageNewProtocolQuery(
+                self._message_protocol_version,
+                supports_rate_select=self._capabilities.get("rate_select", False),
+            ),
             MessagePowerQuery(self._message_protocol_version),
             MessageHumidityQuery(self._message_protocol_version),
             MessageGroupZeroQuery(self._message_protocol_version),
--- a/midealocal/devices/ac/message.py
+++ b/midealocal/devices/ac/message.py
@@ -449,19 +449,6 @@ class MessageToggleDisplay(MessageACBase):
 class MessageNewProtocolQuery(MessageACBase):
     """AC message new protocol query."""
 
-    def __init__(
-        self,
-        protocol_version: int,
-        *,
-        supports_rate_select: bool = False,
-    ) -> None:
-        """Initialize AC message new protocol query.
-
-        `supports_rate_select` gates the rate_select (0x0048) query param on
-        the device having advertised it via the B5 b5_electricity capability
-        (tag 0x0216). Devices that don't report it never answer the query, so
-        it's left out until support is confirmed.
-        """
     _query_params: tuple[int, ...] = (
         NewProtocolTags.indirect_wind,
         NewProtocolTags.breezeless,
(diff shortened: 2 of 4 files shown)


rokam (commented):
It looks good. Is `b5_eletricity` sent by the devices even if it isn't queried?

wuwentao:
thank you very much for your quickly response and PR fix solution.
I have fixed the PR conflicts, if there is any new bug, we can continue improve, thanks!

chemelli74:
@wuwentao there was a question from @rokam unanswered.

Next time please be patient and give the author time to reply before going straight forward and merge.

Thank you

vaughnitem:
> It looks good. Is `b5_eletricity` sent by the devices even if it isn't queried?

"b5_electricity" is published only in supported/newer midea AC units and its absent in older ones

PR #638, opened 2026-08-13
fix(ed): cancel queued tea bar heating after filling

xiaopan007:
## Summary

- send the verified heat-stop field without re-sending a target temperature
- remember a local stop request made while model `63000622` / subtype `395` is dispensing
- if the firmware starts its queued heating phase after dispensing ends, repeat the same verified heat-stop command exactly once
- clear the pending cancellation when a later local start or explicit temperature request supersedes it

## Why

Real-device debug logs showed that the appliance acknowledged a heat-stop request during dispensing but still entered heating after dispensing completed, without a second start command from Home Assistant. Re-sending a target together with the stop could also leave that target queued.

This change does not send an unverified water-dispensing stop command. Dispensing completes, and only the already verified heat-stop field is used. The behavior remains behind the exact model/subtype guard.

## Validation

- ED focused tests: 127 passed
- full test suite after rebasing onto current main: 1546 passed
- `uv run prek run --all-files`: passed
- real device:
  - stopping during dispensing prevented the queued heating phase
  - leaving the cycle on continued into normal heating
  - no repeated stop loop was observed

Ported in parallel to the successor library: wuwentao/midea-lan#5
Related: wuwentao/midea_ac_lan#668

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tea-bar heating cancellation during dispensing.
  * Ensured heating stops reliably when dispensing ends.
  * Prevented unnecessary temperature commands when stopping heat.
  * Allowed new heating requests to correctly override pending cancellations.

* **Tests**
  * Expanded coverage for heating start, stop, cancellation, and retry behavior.

### DIFF
--- a/midealocal/devices/ed/__init__.py
+++ b/midealocal/devices/ed/__init__.py
@@ -144,6 +144,7 @@ class MideaEDDevice(MideaDevice):
             },
         )
         if self._is_tea_bar():
+            self._tea_bar_cancel_after_dispensing = False
             self._attributes.update(
                 {
                     DeviceAttributes.current_temperature: None,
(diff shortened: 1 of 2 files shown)


rokam (approved):
Looks good to me.

PR #661, opened 2026-08-13
fix(c3): eco message index out of bounds

### LINKED ISSUE #606: IndexError in C3ECOBody.__init__ when device responds to ECO set command (model 171H120F)
### Description

Setting `eco_mode` (either `true` or `false`) on a C3-family heat pump (Midea/Clivet
"new energy" protocol, model **171H120F**) crashes the integration with an
`IndexError`. The crash happens on the device's own "set" acknowledgment packet —
not just on activation, so it can occur on *any* eco_mode write (true or false).

### Traceback

```
File ".../midealocal/devices/c3/message.py", line 373, in __init__
    self.eco_function_state = body[data_offset] & 0x01 > 0
IndexError: bytearray index out of range
```

### Root cause

`C3ECOBody.__init__` (`midealocal/devices/c3/message.py`) indexes `body[data_offset]`
without checking that `body` is long enough. This model's "set" acknowledgment
returns a body containing only the body-type marker byte (`07`), with no data byte
at `data_offset` — a valid (if minimal) response the parser should tolerate rather
than crash on.

Confirmed via debug logs — the exact crashing packet:
```
aa0bc3000000000000020729
```
`_message_type: set`, `_body_type: X07`, `_data: '07'` (1 byte only; `data_offset`
expects at least 2).

### Fix

Add a bounds check before indexing:

```python
# Before
self.eco_function_state = body[data_offset] & 0x01 > 0
self.eco_timer_state = body[data_offset] & 0x02 > 0

# After
self.eco_function_state = len(body) > data_offset and body[data_offset] & 0x01 > 0
self.eco_timer_state = len(body) > data_offset and body[data_offset] & 0x02 > 0
```

Verified over several hours on the affected device with this exact patch applied
at runtime: no crashes, all other sensor values remain populated.

### Prior report

Same crash, same model, already reported against the (now inactive) predecessor
repo: georgezhao2010/midea_ac_lan#666 — never picked up there since that repo
isn't maintained anymore. Filing here since this is where the C3 parsing code
now actually lives.

rokam:
Fixes: #606

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of short ECO responses to prevent errors when state information is missing.
  * ECO and timer status are now reported correctly across supported response variations.

* **Tests**
  * Expanded coverage for ECO and timer combinations, including responses with missing state data.

### DIFF
--- a/midealocal/devices/c3/message.py
+++ b/midealocal/devices/c3/message.py
@@ -385,8 +385,10 @@ class C3ECOBody(MessageBody):
     def __init__(self, body: bytearray, data_offset: int = 0) -> None:
         """Initialize C3 ECO message body."""
         super().__init__(body)
-        self.eco_function_state = body[data_offset] & 0x01 > 0
-        self.eco_timer_state = body[data_offset] & 0x02 > 0
+        self.eco_function_state = (
+            len(body) > data_offset and body[data_offset] & 0x01 > 0
+        )
+        self.eco_timer_state = len(body) > data_offset and body[data_offset] & 0x02 > 0
 
 
 class C3DisinfectBody(MessageBody):
--- a/midealocal/devices/c3/message.py
+++ b/midealocal/devices/c3/message.py
@@ -12,6 +12,8 @@ from midealocal.message import (
 )
 
 TEMP_NEG_VALUE = 127
+ECO_FUNCTION_STATE_MASK = 0x01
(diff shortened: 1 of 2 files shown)


rokam:
@coderabbitai review

chemelli74 (approved):
LGTM !

PR #679, opened 2026-08-18
refactor(b8)!: cleanup code

chemelli74:
- remove redundant code

### DIFF
--- a/midealocal/devices/b8/message.py
+++ b/midealocal/devices/b8/message.py
@@ -12,15 +12,6 @@ from midealocal.message import (
 )
 
 
-class B8WorkMode(IntEnum):
-    """Midea B8 work mode."""
-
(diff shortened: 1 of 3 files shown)


midealocal/devices/b8/message.py
```diff
@@ -12,15 +12,6 @@
 )
 
 
-class B8WorkMode(IntEnum):
-    """Midea B8 work mode."""
-
-    CHARGE = 0x01
-    WORK = 0x02
-    STOP = 0x03
-    PAUSE = 0x1B
```
chemelli74, line 21: This is different than the value in `B8WorkStatus`; which one is the correct one ?

rokam: WorkMode is a set attr and WorkStatus is readonly

PR #684, opened 2026-08-19
fix(cd): support Kosner 2530001N and sync power with mode selection

### LINKED ISSUE #576: [water heater] Kosner 2530001N support and None mode does not power off the device
Hi, first of all, thank you for the amazing work on this project.

I have a heat pump water heater from the Spanish company [Kosner](https://www.kosner.es/), the **KCA V4.2 100**, with Part Number **2530001N**. 
Kosner seems to be using Midea hardware, at least for their remote control interface, since the ACs they distribute also work with NetHome Plus app (and therefore this integration).

I have found that this particular device is very similar to the **RSJRAC07**. For it to properly work I added it to the list of devices that use the new protocol as well as to the model lists that require the special temperature handling. I've made this change and it works properly now:

```
/midealocal/devices/cd/__init__.py

@@ -220,6 +220,7 @@ class MideaCDDevice(MideaDevice):
                     "RSJRAC01",
                     "RSJRAC06",
                     "RSJRAC07",
+                    "2530001N",
                 }
                 return_value = LuaProtocol.new if check_device else LuaProtocol.old
         if isinstance(value, bool | int):
@@ -288,10 +289,10 @@ class MideaCDDevice(MideaDevice):
                     parsed = self._value_to_temperature(
                         raw_value,
                         force_fahrenheit=(
-                            self.model in ["RSJRAC06", "RSJRAC07"] and is_outdoor_temp
+                            self.model in ["RSJRAC06", "RSJRAC07", "2530001N"] and is_outdoor_temp
                         ),
                         force_old=(
-                            self.model in ["RSJRAC06", "RSJRAC07"] and is_current_temp
+                            self.model in ["RSJRAC06", "RSJRAC07", "2530001N"] and is_current_temp
                         ),
                     )
                     # Defensive: ignore invalid zeros for min/max/target/current
```

However, I found another issue with the current implementation. When selecting the `None` mode, nothing actually happens on the device. 

Since the Home Assistant `midea_ac_lan` card does not expose a separate "OFF" button, I would expect that selecting None would power off the device, just like with the air conditioners.

For that to happen, selecting `None` should also set the `Power` attribute to `False`. Likewise, if the device is off, selecting any other mode should set `Power` back to `True` as well as selecting the mode.

I'm not sure whether this behavior is specific to my device or applies to all CD devices. I'm also not sure whether this logic belongs in this library or should instead be implemented in midea_ac_lan.

As a quick test, I made the following change, and it works correctly with my device:

```
/midealocal/devices/cd/__init__.py

@@ -484,14 +485,12 @@ class MideaCDDevice(MideaDevice):
                     "_modes",
                     str(value),
                 )
-                if mode_key is None:
-                    _LOGGER.warning(
-                        "[%s] Invalid mode value: %s, not sending command",
-                        self.device_id,
-                        value,
-                    )
-                    return  # Don't send invalid mode
-                message.mode = mode_key
+
+                if mode_key is None or mode_key == 0x00:
+                    message.power = False
+                else:
+                    message.power = True
+                    message.mode = mode_key
```

However, I don't think this is the cleanest or most generic solution, so I'd appreciate your opinion on the proper way to handle this.

Again, thanks a lot for your work.

chemelli74:
fixes #576

## Summary by CodeRabbit

* **New Features**
  * Improved automatic protocol selection for newer CD device models.
  * Added model-specific temperature handling for affected devices.
  * Mode changes now automatically power devices on or off as appropriate.

* **Bug Fixes**
  * Improved temperature conversion and fallback behavior.
  * Improved vacation-day handling and Lua protocol customization.
  * Expanded coverage for model-specific and fallback scenarios.

### DIFF
--- a/midealocal/devices/cd/__init__.py
+++ b/midealocal/devices/cd/__init__.py
@@ -109,6 +109,15 @@ class MideaCDDevice(MideaDevice):
         0x05: "vacation",
     }
     _vacation_mode_key: ClassVar[int] = 0x05
+    # models that report over the new (raw °C) Lua protocol
+    _new_protocol_models: ClassVar[frozenset[str]] = frozenset(
+        {"RSJRAC01", "RSJRAC06", "RSJRAC07", "2530001N"},
+    )
+    # new-protocol models whose outdoor/current temps still need the old
+    # fahrenheit/°C decoding quirk (RSJRAC01 does not need this)
+    _forced_temperature_models: ClassVar[frozenset[str]] = frozenset(
+        {"RSJRAC06", "RSJRAC07", "2530001N"},
+    )
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


rokam (approved):
LGTM

PR #686, opened 2026-08-19
fix(fa): add opt-in mode override for fan devices

### LINKED ISSUE #504: FA fan: 'Normal' mode is unsettable — bit 0x20 selects it and the setter's 0x1E mask can't emit it
Device: **0xFA tower fan, model `56000211`, subtype `0`, protocol V3**.

Three of the four modes this fan implements can be set from `midea-local`.
**Normal cannot be set at all** — and the reason turns out to be a bit the mode
setter structurally cannot produce.

## What the device actually does

State byte is `body[4]` in `FAGeneralMessageBody`: `bit0` = power,
`bits1-4` = mode+1. Captured frames (debug logging, real device):

| App mode | state byte | parsed mode index | `_modes[i]` label |
|---|---|---|---|
| Normal | **`0x29`** | 3 | `Comfort` |
| Nature | `0x03` | 1 | `Natural` |
| Sleep | `0x07` | 2 | `Sleep` |
| Smart | `0x1B` | 12 | `Smart` |

Normal reads back as `0x29` = power `0x01` | nibble 4 (`0x08`) | **`0x20`**.

The setter is:

```python
if self.mode is not None:
    _body_return[3] = 1 | (((self.mode + 1) << 1) & 0x1E)
```

The `0x1E` mask means `0x20` can never be emitted, so requesting mode index 3
sends `0x09`. The fan **ignores it** — captured response frames show the device
echoing its previous mode back unchanged. Requesting index 0 (`0x03`) doesn't
select Normal either; that's Nature.

## `0x20` is the Normal *selector*, not an extra flag

I tested this rather than assuming it. Forcing `0x20` on **every** mode set made
the fan go to Normal **whichever mode was requested** — Nature, Sleep and Smart
all landed on Normal. So on this model:

- `0x20` set → Normal (overrides the nibble entirely)
- `0x20` clear → the nibble picks Nature / Sleep / Smart

Applying the bit unconditionally therefore silently breaks the other three
modes. It has to be gated to the Normal case.

## Two smaller things this implies

- `_modes[3]` is labelled `Comfort`, but on this model device value 3 **is**
  Normal. Read-back and write are at least self-consistent, so a user selects
  "Comfort" in HA to get Normal — confusing, but I don't know whether the
  generic table is wrong or just doesn't fit this model.
- This fan reports `fan_speed` 1–12 while `_default_speed_count = 3`, so HA
  publishes `percentage` up to 400 (outside its legal 0–100) unless
  `customize` sets `speed_count: 12`. That part already has an extension point
  and works fine — mentioning it only because it's the same device.

## Proposed fix — opt-in, zero effect on other devices

I have exactly **one** FA device, so I don't want to guess what `0x20` means
across the family or hard-code a mode index into the shared setter. The
existing `set_customize` hook already carries per-device quirks (`speed_count`),
so the smallest safe shape is another opt-in key there, e.g.:

```jsonc
{"speed_count": 12, "normal_mode_bit": 32, "normal_mode_index": 3}
```

…applied only when the requested mode index matches, leaving every device that
doesn't set it on exactly today's code path. A more general alternative would be
an explicit override table (`{"mode_set_overrides": {"3": 41}}`) that lets any
device with a quirky mode encoding be fixed from config without a code change.

I'm running the gated version locally and all four modes now set correctly. I'm
happy to send it as a PR — I'd rather you pick the API shape first, since you
know whether other FA models share this encoding. Just say which you'd prefer
(or that you'd rather handle it differently) and I'll open it.

Frames, byte-level analysis and the local patch:
https://github.com/vyakunin/smart_home (not public — I can paste anything useful
here on request).

Related: #503 (the short-frame `IndexError`, hit on the same device).

chemelli74:
fixes #504

### DIFF
--- a/midealocal/devices/fa/__init__.py
+++ b/midealocal/devices/fa/__init__.py
@@ -104,6 +104,7 @@ class MideaFADevice(MideaDevice):
         )
         self._default_speed_count = 3
         self._speed_count: int = self._default_speed_count
+        self._mode_set_overrides: dict[int, int] = {}
         self.set_customize(customize)
 
     @property
@@ -308,6 +309,7 @@ class MideaFADevice(MideaDevice):
             if value in MideaFADevice._modes:
                 message = MessageSet(self._message_protocol_version, self.subtype)
                 message.mode = MideaFADevice._modes.index(str(value))
+                message.mode_set_overrides = self._mode_set_overrides
         elif not (attr == DeviceAttributes.fan_speed and value == 0):
             message = MessageSet(self._message_protocol_version, self.subtype)
             setattr(message, str(attr), value)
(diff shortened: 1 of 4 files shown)


rokam (approved):
LGTM

PR #696, opened 2026-08-21
fix(b8): handle unknown error sub-codes in MessageB8GenericBody

chemelli74:
Every other coded field in MessageB8GenericBody falls back on ValueError; error_desc's three enum lookups didn't, so an unrecognized error sub-code dropped the entire status update. 

Added the same try/except pattern.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid error descriptions for `CAN_FIX`, `REBOOT`, and `WARNING` error types.
  * Invalid values now safely default to the appropriate “no error” state instead of causing failures.


rokam (approved):
LGTM

PR #698, opened 2026-08-21
fix(fc): correct FCNotifyMessageBody standby bitmask check

chemelli74:
FCNotifyMessageBody.standby had its bitmask and comparison value swapped (body[27] & 0x14 == 0xFF), making it tautologically always False. 
Fixed to match the correct sibling check.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved standby-state detection for FC device notifications.
  * Standby status is now reported accurately when the device sends the expected standby signal.
  * Existing handling for incomplete notification data remains unchanged, preserving safe fallback behavior.

### DIFF
--- a/midealocal/devices/fc/message.py
+++ b/midealocal/devices/fc/message.py
@@ -231,9 +231,7 @@ class FCNotifyMessageBody(MessageBody):
             self.tvoc = body[15]
         self.anion = (body[10] & 0x20 > 0) if len(body) > ANION_NOTIFY_BYTE else False
         self.standby = (
-            ((body[27] & 0xFF) == STANDBY_VALUE)
-            if len(body) > STANDBY_NOTIFY_BYTE
-            else False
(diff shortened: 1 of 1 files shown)


midealocal/devices/fc/message.py
```diff
@@ -231,7 +231,7 @@ def __init__(self, body: bytearray) -> None:
             self.tvoc = body[15]
         self.anion = (body[10] & 0x20 > 0) if len(body) > ANION_NOTIFY_BYTE else False
         self.standby = (
-            (body[27] & 0x14 == MAX_BYTE_VALUE)
+            ((body[27] & 0xFF) == STANDBY_VALUE)
```
rokam:
```suggestion
            (body[27] == STANDBY_VALUE)
```

The body[27] is already one byte. This & is unnecessary.

rokam (dismissed):

midealocal/devices/fc/message.py
chemelli74: Code updated

rokam (approved):

PR #699, opened 2026-08-21
fix(e8): guard E8MessageBody against short payloads

chemelli74:
~~Same class of bug: E8MessageBody reads up to body[43] but was only guarded for length >6.~~
~~Raised the guard to >43.~~

Update: use parser to avoid future issues.


rokam (commented):
What do you think about creating a helper inside MessageBody to get the value of a byte? 

`def get_value(int bit, default=None, lenght=1)`

In this method it will check if body has the enought lenght and return it. Maybe also with lenght > 1, you can do the little_endian or big_endian to concatenate multiple bytes.

This will prevent future issues like this.

rokam (approved):
This does work for now. But I think we should migrate to bodyparsers to avoid additional issues.

PR #700, opened 2026-08-21
fix(ca): guard CAGeneralMessageBody against short payloads

chemelli74:
CAGeneralMessageBody reads up to body[24] but the length guard only required >20; a 21–24 byte body raised IndexError and dropped the whole update.
Raised the guard to >24.

Update: use parser


rokam (approved):
This does work for now. But I think we should migrate to bodyparsers to avoid additional issues.

PR #717, opened 2026-08-24
fix(cloud): use the v2 getToken endpoint for the Meiju cloud

wawnnzxd:
Requested by @chemelli74 in wuwentao/midea-lan#44 — same change, ported to this repo.

## The problem

The Meiju cloud retired `/v1/iot/secure/getToken`, which is what
`MideaCloud.get_cloud_keys()` calls. The gateway now answers:

```json
{"code": 40404, "msg": "the access address does not exist"}
```

So no keys come back at all and **every V3 device fails to authenticate** with
"Can't get available token from Midea server". `MeijuCloud` does not override
`get_cloud_keys()`, so it inherits the dead v1 path.

## How the replacement was pinned down

The v2 path is not documented anywhere, so I probed it with the gateway's own
error codes, always alongside a known-good and a known-bad control:

| request | response | reading |
|---|---|---|
| `/v1/iot/secure/getToken` | `40404` the access address does not exist | path is gone |
| `/v2/iot/secure/getToken`, v1 payload | `1002` none parameter is found | **path exists**, payload is wrong |
| `/v2/...` + `homegroupId` | `1002` | still missing something |
| `/v2/...` + `homegroupId`, `applianceCodes` as a list | `0` + `tokenlist` | correct |
| `/v3/iot/secure/getToken` | `40404` | v2 is the current one |

`40404` (path missing) versus `1002` (path present, parameters wrong) is what
separates "wrong URL" from "wrong body" — without that contrast the v2 endpoint
looks just as dead as v1.

Two things changed from v1:

1. `homegroupId` is now required — hence the walk over `list_home()`.
2. `applianceCodes` must be a **list**; the plain string v1 accepted returns `1002`.

## The change

`MeijuCloud.get_cloud_keys()` walks the account's homes, queries v2, and **falls
back to the inherited v1 implementation when v2 yields nothing** — so this is a
no-op for any cloud or account the old endpoint still serves.

The debug line logs only the token-entry count. The payload carries `token` and
`key` material and must not be written to logs.

## Verification

Real hardware, on a Midea AC that could not authenticate before this change:
`get_cloud_keys()` now returns both method 1 and method 2 keys and the device
comes up.

`pytest ./tests/` — 1619 passed. `prek run --all-files` — clean.

The tests pin the **outbound request contract**, not just the responses: exact
endpoint, exact `homegroupId` per home, the method-specific `udpid`, and the
list-valued `applianceCodes`, across the v2 path, the v1 fallback and the
no-home path. I checked the assertions actually bite by mutating the
implementation five ways — each one turns the suite red:

| mutation | result |
|---|---|
| endpoint reverted to `/v1/iot/secure/getToken` | 2 failed |
| `applianceCodes` back to `str(appliance_id)` | 2 failed |
| `homegroupId` dropped from the payload | 2 failed |
| `homegroupId` hard-coded to `"1"` | 1 failed |
| `udpid` always computed with method 1 | 2 failed |

## Note on a separate issue

The inherited `MideaCloud.get_cloud_keys()` v1 path logs its full response
(`"Response from get_keys() ... %s", response`), which is the same token/key
material. That is present in this repo too. I kept it out of this PR to stay
scoped; it is filed upstream as wuwentao/midea-lan#45. Happy to send a small
follow-up PR here if you want it fixed.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,9 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+
+language: "en-US"
+
+reviews:
+  # Keep the CodeRabbit summary, but put it in the walkthrough
+  # instead of modifying the PR description.
+  high_level_summary: true
+  high_level_summary_in_walkthrough: true
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,7 +54,7 @@ repos:
           - pycryptodome
           - platformdirs
   - repo: https://github.com/PyCQA/pylint
-    rev: v4.0.7
+    rev: v4.0.8
     hooks:
       - id: pylint
         name: pylint
--- a/midealocal/cloud.py
+++ b/midealocal/cloud.py
@@ -6,7 +6,9 @@ import logging
 import re
 import time
 from asyncio import Lock
+from collections.abc import Mapping
 from datetime import UTC, datetime
+from hashlib import md5
 from http import HTTPStatus
 from pathlib import PurePosixPath, PureWindowsPath
 from secrets import token_hex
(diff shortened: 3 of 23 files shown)


rokam (dismissed):
Good enought for me.

rokam (approved):

PR #718, opened 2026-08-24
fix(cloud): keep getToken credentials out of the debug log

wawnnzxd:
Requested by @chemelli74 in wuwentao/midea-lan#44, and it also covers the
credential-logging risk CodeRabbit flagged on #717.

## Two separate leaks

**1. `_redact_data()` has no rule for `token` / `key`.**

A getToken response goes through `_api_request()`, which logs
`_redact_data(str(raw))`. None of the existing patterns target the JSON
credential fields — the only thing that touches them is the credit-card pattern
`\b(?:\d[ -]*?){13,19}\b` happening to match digit runs inside the hex value.
That produces output that *looks* redacted while leaving most of the secret
readable:

```
"token":"AABBCCDDEEFF00112**********78899AABBCCDDEEFF00112**********78899"
"key":"00112**********78899**********55667**********23344**********0011"
```

After this change:

```
"token":"AABBC***********************************************************"
"key":"00112***********************************************************"
```

The new rule runs *before* the generic patterns, so they cannot chew the value
up first, and reuses `_mask_token()` so the output matches the rest of the
module. It tolerates the escaped quotes you get from `str(bytes)` rendering.

**2. `MideaCloud.get_cloud_keys()` logs its full parsed response.**

```python
_LOGGER.debug(
    "Response from get_keys() for appliance_id %s with method %s: %s",
    appliance_id, method, response,
)
```

That is the same `tokenlist`, and unlike the `_api_request()` line it is not
passed through `_redact_data()` at all. Now logs the entry count.

## Tests

Asserting only "the raw token is absent from the redacted string" **passes
without the fix**, because the credit-card pattern already breaks the string up.
So the test asserts the exact masked form instead, and a second test drives a
real `get_cloud_keys()` call and checks no emitted record contains the token or
the key.

I verified the tests actually bite by mutating the implementation three ways:

| mutation | result |
|---|---|
| drop the `token`/`key` redaction | 3 failed |
| `get_cloud_keys()` logs `response` again | 1 failed |
| redaction rule matches `token` but not `key` | 2 failed |

`pytest ./tests/` — 1620 passed. `prek run --all-files` — clean.

Independent of #717 — this branches from `main` and touches different regions of
`cloud.py`, so the two can merge in either order. The equivalent of leak 2 is
filed upstream as wuwentao/midea-lan#45; leak 1 is present there too and I am
happy to send the same change to that repo.

### DIFF
--- a/midealocal/cloud.py
+++ b/midealocal/cloud.py
@@ -123,8 +123,19 @@ def _mask_token(token: str) -> str:
     return visible + block * max(0, len(token) - len(visible))
 
 
+# ``"token": "..."`` / ``"key": "..."`` as they appear in a getToken response,
+# tolerating escaped quotes from ``str(bytes)`` rendering.
+_CREDENTIAL_FIELD = re.compile(
+    r'(\\?["\'](?:token|key)\\?["\']\s*:\s*\\?["\'])([^"\'\\]+)',
+    re.IGNORECASE,
+)
+
+
 def _redact_data(data: str) -> str:
     """Redact sensitive data."""
+    # Do this first: the generic patterns below only chew up parts of a token,
+    # which leaves most of the credential readable and looks redacted.
+    data = _CREDENTIAL_FIELD.sub(lambda m: m.group(1) + _mask_token(m.group(2)), data)
     patterns = [
         # Email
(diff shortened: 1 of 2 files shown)


rokam (approved):
LGTM

PR #719, opened 2026-08-24
fix(device): follow up of #690

chemelli74:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved appliance connection recovery after communication errors.
  * Unsupported protocols now trigger a clean socket close and reconnection.
  * Appliance detection is correctly re-enabled for subsequent connection attempts.

* **Tests**
  * Added coverage for socket recovery and reconnection after unsupported protocols.


rokam (approved):
LGTM

PR #722, opened 2026-08-26
fix(e6): heating_modes must be named preset_modes for compatibility

rokam:
Water_heater uses `preset_modes` and not `heating_modes` to set in the device. This should be unified for better compatibility

## Summary by CodeRabbit

* **New Features**
  * Added a `preset_modes` property for accessing available preset heating modes.

* **Deprecation**
  * The `heating_modes` property is now deprecated and may display a warning. Use `preset_modes` instead.

### DIFF
--- a/midealocal/devices/e6/__init__.py
+++ b/midealocal/devices/e6/__init__.py
@@ -5,6 +5,8 @@ import logging
 from enum import StrEnum
 from typing import Any, ClassVar, Unpack
 
+from typing_extensions import deprecated
+
 from midealocal.const import DeviceType
 from midealocal.device import MideaDevice, MideaDeviceInitKwargs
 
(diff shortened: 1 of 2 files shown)


midealocal/devices/e6/__init__.py
```diff
@@ -103,9 +105,15 @@ def set_attribute(self, attr: str, value: bool | float | str) -> None:
             self.build_send(message)
 
     @property
+    @deprecated("Use preset_modes instead.")
```
chemelli74, line 108: Do we really need deprecation ? As we never released this in HomeAssistant I think we can go as far as just dropping it

chemelli74 (changes requested):

midealocal/devices/e6/__init__.py
rokam: I think we need to do the right thing, no matter how many projects use this. That way, anyone using the lib will receive the correct flow of info.

chemelli74: Agreed, and announcing a change as breaking is exactly what this means. ;-)

chemelli74 (approved):

PR #732, opened 2026-08-27
fix(fb): align to climate platform

chemelli74:


midealocal/devices/fb/__init__.py
```diff
@@ -92,12 +95,14 @@ def set_attribute(self, attr: str, value: bool | float | str) -> None:
...
         target_temperature: float,
-        mode: int | None,  # noqa: ARG002
+        mode: int | None,
         zone: int | None = None,  # noqa: ARG002
     ) -> None:
         """Midea FB device set target temperature."""
         message = MessageSet(self._message_protocol_version, self.subtype)
         setattr(message, DeviceAttributes.target_temperature, target_temperature)
+        if mode is not None:
+            message.power = True
```
rokam: If `mode == 0` `self.power` should also be set to `True`?

rokam (changes requested):

midealocal/devices/fb/__init__.py
chemelli74:
```
Python 3.14.5 (main, Jun  2 2026, 22:25:17) [Clang 22.1.3 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
Ctrl click to launch VS Code Native REPL
>>> mode = 0
>>> if mode is not None:
...     print("OK")
... else:
...     print("Mode is None")
... 
OK
```

rokam: I think I didn't make myself clear. mode = 0 doesn´t exists. So, if the person set it to zero I don't think power should be set to True.

chemelli74: ah ok, sorry

chemelli74: Fixed

rokam (approved):

PR #735, opened 2026-08-28
feat(cd): support extended water heater protocol

FrancYescO:
## Summary

- add CD extended status, B1 capability queries, B0 maintenance writes, and daily/weekly schedule payloads
- expose reported limits and feature flags without a model allowlist
- support extended disinfection, maximum-temperature, and schedule-mode controls
- keep automatic disinfection read-only because the official app does not expose it as a toggle
- preserve the last valid maintenance-reminder value when a B1 response omits that TLV

## Validation

- `ruff check` and `ruff format`
- mypy, pylint, codespell, pyupgrade, and prettier pre-commit hooks
- 164 focused CD tests
- verified on a real CD water heater through Home Assistant 2026.8.3 over multiple polling cycles

Companion integration work is tracked separately in `wuwentao/midea_ac_lan`.

## Summary by CodeRabbit

* **New Features**
  * Added support for extended water-heater capabilities and operating modes.
  * Added temperature limits, hot-water status, heating options, demand-response settings, and capability indicators.
  * Added maintenance, disinfection, sterilization-temperature, daily, weekly, holiday, and schedule-mode controls.
  * Added holiday, hybrid, heat-recovery, boost, silent, and smart operating features.

* **Bug Fixes**
  * Preserved capabilities when responses omit previously reported fields.
  * Enforced temperature limits and rejected malformed schedules before submission.
  * Improved automatic disinfection status reporting and operating-mode selection.

### DIFF
--- a/midealocal/devices/cd/__init__.py
+++ b/midealocal/devices/cd/__init__.py
@@ -488,7 +488,7 @@ class MideaCDDevice(MideaDevice):
             DeviceAttributes.condenser_temperature,
             DeviceAttributes.compressor_temperature,
         ]
-        return self.update_attributes_from_message(
+        new_status = self.update_attributes_from_message(
             message,
             {
                 # Only update mode on a recognised mode key, to prevent
@@ -526,6 +526,18 @@ class MideaCDDevice(MideaDevice):
                 ),
             },
         )
+        # Extended heaters report the stable maintenance state through the
+        # canonical 0x40/B1 flag; their legacy 0x80 bit can flap independently.
+        # Mirror the canonical value so existing HA entity IDs remain stable.
+        maintenance_reminder = new_status.get(
+            str(DeviceAttributes.maintenance_reminder),
+        )
+        if isinstance(maintenance_reminder, bool) and self._is_extended_water_heater(
+            message,
+        ):
+            self._attributes[DeviceAttributes.maintain_warn] = maintenance_reminder
+            new_status[str(DeviceAttributes.maintain_warn)] = maintenance_reminder
+        return new_status
 
     def _set_maintenance_reminder(self, value: bool) -> None:
         """Send the dedicated B0 maintenance flag without changing other flags."""
--- a/midealocal/devices/cd/__init__.py
+++ b/midealocal/devices/cd/__init__.py
@@ -4,7 +4,7 @@ import json
 import logging
 from collections.abc import Callable
 from enum import IntEnum, StrEnum
-from typing import Any, ClassVar, Unpack
+from typing import Any, ClassVar, Unpack, cast
 
 from midealocal.const import DeviceType
 from midealocal.device import (
@@ -15,6 +15,7 @@ from midealocal.device import (
 )
 
 from .message import (
+    DailyTimerSchedule,
     MessageCDBase,
     MessageCDResponse,
     MessageQuery,
@@ -26,6 +27,7 @@ from .message import (
     MessageSetMaintenance,
     MessageSetSterilize,
     MessageSetWeekly,
+    WeeklySchedule,
 )
 
 _LOGGER = logging.getLogger(__name__)
@@ -369,7 +371,9 @@ class MideaCDDevice(MideaDevice):
             DeviceAttributes.support_tou,
         )
 
-        def _source_is_extended(source: object) -> bool:
+        def _source_is_extended(source: object | None) -> bool:
+            if source is None:
+                return False
             if _value(source, DeviceAttributes.new_version_water_heater) is True:
                 return True
             if isinstance(_value(source, DeviceAttributes.b0_reserved_flags), int):
@@ -382,8 +386,6 @@ class MideaCDDevice(MideaDevice):
                 return True
             return any(_value(source, attribute) is True for attribute in capabilities)
 
-        if message is None:
-            return _source_is_extended(self._attributes)
         return _source_is_extended(message) or _source_is_extended(self._attributes)
 
     def _mode_map(self, message: object | None = None) -> dict[int, str]:
@@ -398,6 +400,8 @@ class MideaCDDevice(MideaDevice):
         """Build a mode translator using the capabilities in this response."""
 
         def _translate(value: int) -> Any:  # noqa: ANN401
+            # Extended mode 0x05 normally means Heat Pump, but the dedicated
+            # vacation flag gives the same value Vacation semantics.
             if self._is_extended_water_heater(message) and getattr(
                 message,
                 DeviceAttributes.vacation_mode,
@@ -407,6 +411,13 @@ class MideaCDDevice(MideaDevice):
             return self._mode_map(message).get(value, SKIP_ATTRIBUTE)
 
         return _translate
+
+    def _mode_key(self, value: str) -> int | None:
+        """Return the protocol key from the capability-selected mode map."""
+        return next(
+            (key for key, name in self._mode_map().items() if name == value),
+            None,
+        )
 
     def build_query(self) -> list[MessageCDBase]:
         """Midea CD device build query."""
@@ -667,6 +678,11 @@ class MideaCDDevice(MideaDevice):
         else:
             if "amount" in value and not cls._is_integer_compatible(value["amount"]):
                 return False
+            if any(
+                flag in value and not isinstance(value[flag], bool)
+                for flag in ("single_timer_on", "single_timer_off")
+            ):
+                return False
             timer_groups = [value.get("timers", [])]
             numeric_fields = (
                 "openhour",
@@ -680,6 +696,7 @@ class MideaCDDevice(MideaDevice):
             isinstance(timers, list)
             and all(
                 isinstance(timer, dict)
+                and ("effect" not in timer or isinstance(timer["effect"], bool))
                 and all(
                     field not in timer or cls._is_integer_compatible(timer[field])
                     for field in numeric_fields
@@ -732,11 +749,11 @@ class MideaCDDevice(MideaDevice):
                 return
             if attr == DeviceAttributes.weekly_schedule:
                 weekly = MessageSetWeekly(self._message_protocol_version)
-                weekly.weekly_schedule = value
+                weekly.weekly_schedule = cast("WeeklySchedule", value)
                 self.build_send(weekly)
             else:
                 daily = MessageSetDaily(self._message_protocol_version)
-                daily.daily_timer_schedule = value
+                daily.daily_timer_schedule = cast("DailyTimerSchedule", value)
                 self.build_send(daily)
             return
 
@@ -845,14 +862,7 @@ class MideaCDDevice(MideaDevice):
                 # Fall back to 0x00 (no explicit operating mode).
                 message.mode = 0x00
             else:
-                mode_key = next(
-                    (
-                        key
-                        for key, name in self._mode_map().items()
-                        if name == str(current_mode)
-                    ),
-                    None,
-                )
+                mode_key = self._mode_key(str(current_mode))
                 message.mode = mode_key if mode_key is not None else 0x00
 
             # Update based on attribute being set
@@ -865,14 +875,7 @@ class MideaCDDevice(MideaDevice):
                         self.device_id,
                     )
                     return
-                mode_key = next(
-                    (
-                        key
-                        for key, name in self._mode_map().items()
-                        if name == str(value)
-                    ),
-                    None,
-                )
+                mode_key = self._mode_key(str(value))
                 if mode_key is None:
                     _LOGGER.warning(
                         "[%s] Invalid mode value: %s, not sending command",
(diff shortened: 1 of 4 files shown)


midealocal/devices/cd/__init__.py
```diff
@@ -109,6 +145,16 @@ class MideaCDDevice(MideaDevice):
...
+    _extended_modes: ClassVar[dict[int, str]] = {
+        0x00: "none",
+        0x01: "energy_save",
+        0x02: "hybrid",
+        0x03: "e_heater",
+        0x04: "smart",
+        0x05: "heat_pump",
+        0x09: "boost",
+        0x0A: "silent",
+    }
```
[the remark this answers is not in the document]
FrancYescO: Fixed in 69462a9. The vacation guard now resolves the vacation label from the base mode map, and a regression test verifies that heat_pump sends protocol mode 0x05.

tests/devices/cd/device_cd_test.py
```diff
@@ -744,3 +767,134 @@ def test_auto_lua_protocol_resolves_new(self, model: str) -> None:
...
+    def test_schedule_writes_require_support_and_mapping(self) -> None:
+        """Weekly and daily mappings use their dedicated message types."""
+        with patch.object(self.device, "build_send") as mock_send:
+            self.device.set_attribute(DeviceAttributes.weekly_schedule.value, {})
+            assert isinstance(mock_send.call_args.args[0], MessageSetWeekly)
+            self.device.set_attribute(DeviceAttributes.daily_timer_schedule.value, {})
+            assert isinstance(mock_send.call_args.args[0], MessageSetDaily)
+            mock_send.reset_mock()
+            self.device.set_attribute(DeviceAttributes.weekly_schedule.value, True)
+            mock_send.assert_not_called()
```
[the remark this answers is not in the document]
FrancYescO: Fixed in 69462a9. Daily and weekly nested timer collections and entries are validated before message construction; malformed collections and entries are covered by regression tests.

tests/devices/cd/device_cd_test.py
```diff
@@ -744,3 +767,134 @@ def test_auto_lua_protocol_resolves_new(self, model: str) -> None:
...
+
+    def test_scalar_controls_reject_mappings(self) -> None:
+        """Ordinary scalar controls never accept schedule-like mappings."""
+        with patch.object(self.device, "build_send") as mock_send:
+            self.device.set_attribute(DeviceAttributes.power.value, {})
+        mock_send.assert_not_called()
```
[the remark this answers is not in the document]
FrancYescO: Fixed in 69462a9. Mapping values are rejected before maintenance B0 writes, with regression coverage for both maintenance attributes.

midealocal/devices/cd/__init__.py
```diff
@@ -375,27 +526,157 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+        """Return whether a schedule contains safe timer collections."""
+        timer_groups = (
+            [value.get(day, []) for day in range(7)]
+            if attr == DeviceAttributes.weekly_schedule
+            else [value.get("timers", [])]
+        )
+        return all(
+            isinstance(timers, list)
+            and all(isinstance(timer, dict) for timer in timers)
+            for timers in timer_groups
```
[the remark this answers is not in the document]
FrancYescO: Fixed in 24239f5. Daily/weekly integer-encoded fields are validated before message construction, including daily `amount`; regression coverage includes `None` numeric values on both encoder paths.

midealocal/devices/cd/__init__.py
```diff
@@ -375,27 +526,157 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+        )
+        self.build_send(message)
 
-    def set_attribute(self, attr: str, value: bool | float | str) -> None:
+    @staticmethod
+    def _has_valid_timer_data(attr: str, value: dict[Any, Any]) -> bool:
+        """Return whether a schedule contains safe timer collections."""
+        timer_groups = (
+            [value.get(day, []) for day in range(7)]
+            if attr == DeviceAttributes.weekly_schedule
```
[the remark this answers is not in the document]
FrancYescO: Fixed in 24239f5. Weekly schedules now reject non-integer/out-of-range day keys before message construction; regression coverage includes the JSON-style string key `"0"`.

midealocal/devices/cd/__init__.py
```diff
@@ -375,27 +526,189 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+        self.build_send(message)
+
+    @staticmethod
+    def _is_integer_compatible(value: Any) -> bool:  # noqa: ANN401
+        """Return whether a schedule value can be encoded as an integer."""
+        try:
+            int(value)
+        except (TypeError, ValueError, OverflowError):
+            return False
+        return True
```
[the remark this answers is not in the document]
FrancYescO: Fixed in d4b3422. Boolean values are rejected before integer conversion for daily amounts and all timer fields; regressions cover daily `true` and weekly `false`. The equivalent fix is included in midea-lan #61 as 6b149b1.

midealocal/devices/cd/__init__.py
```diff
@@ -245,19 +324,100 @@ def temperature_step(self) -> float | None:
...
+            DeviceAttributes.support_silent_mode,
+            DeviceAttributes.support_remaining_hot_water,
+            DeviceAttributes.support_electric_mode,
+            DeviceAttributes.support_auto_disinfect,
+            DeviceAttributes.support_force_e_heating,
+            DeviceAttributes.support_tou,
+        )
+
+        def _source_is_extended(source: object) -> bool:
+            if _value(source, DeviceAttributes.new_version_water_heater) is True:
```
rokam, line 377:
```suggestion
            if source is None:
                return False
            if _value(source, DeviceAttributes.new_version_water_heater) is True:
```

midealocal/devices/cd/__init__.py
```diff
@@ -245,19 +324,100 @@ def temperature_step(self) -> float | None:
...
+                isinstance(_value(source, attribute), int | float)
+                and _value(source, attribute) > 0
+                for attribute in limits
+            ):
+                return True
+            return any(_value(source, attribute) is True for attribute in capabilities)
+
+        if message is None:
+            return _source_is_extended(self._attributes)
+        return _source_is_extended(message) or _source_is_extended(self._attributes)
```
rokam:
```suggestion
        return _source_is_extended(message) or _source_is_extended(self._attributes)
```

midealocal/devices/cd/__init__.py
```diff
@@ -245,19 +324,100 @@ def temperature_step(self) -> float | None:
...
+    def _make_mode_translator(self, message: object) -> Callable[[int], Any]:
+        """Build a mode translator using the capabilities in this response."""
+
+        def _translate(value: int) -> Any:  # noqa: ANN401
+            if self._is_extended_water_heater(message) and getattr(
+                message,
+                DeviceAttributes.vacation_mode,
+                False,
+            ):
+                return self._modes[self._vacation_mode_key]
```
rokam, line 410: I don't get this part.

If `is_extended_water_heater` why you set the `self.mode` to the `_vacation_mode_key` that is available only for non `is_extended_water_heater`

midealocal/devices/cd/__init__.py
```diff
@@ -375,27 +560,191 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+        if isinstance(value, bool):
+            return False
+        try:
+            int(value)
+        except (TypeError, ValueError, OverflowError):
+            return False
+        return True
+
+    @classmethod
+    def _has_valid_timer_data(cls, attr: str, value: dict[Any, Any]) -> bool:
```
rokam, line 667: If you set the correct type for the value, a lot of checks can be dropped.
```suggestion
    def _has_valid_timer_data(cls, attr: str, value: dict[int|str, list[dict[str, str|int|float]]]) -> bool:
```

midealocal/devices/cd/__init__.py
```diff
@@ -375,27 +560,191 @@ def process_message(self, msg: bytes) -> dict[str, Any]:
...
+                )
+                for timer in timers
+            )
+            for timers in timer_groups
+        )
+
+    def set_attribute(  # noqa: C901, PLR0911
+        self,
+        attr: str,
+        value: bool | float | str | dict[Any, Any],
```
rokam, line 712: This dict can also be better typed. I do like: `dict[int|str,list[dict[str,int]]]`

But I don´t know if this is enough.

midealocal/devices/cd/__init__.py
```diff
@@ -476,25 +845,33 @@ def set_attribute(self, attr: str, value: bool | float | str) -> None:
...
-                    "_modes",
-                    str(current_mode),
+                mode_key = next(
+                    (
+                        key
+                        for key, name in self._mode_map().items()
+                        if name == str(current_mode)
+                    ),
+                    None,
                 )
```
rokam: Why did you change this?

midealocal/devices/cd/__init__.py
```diff
@@ -476,25 +845,33 @@ def set_attribute(self, attr: str, value: bool | float | str) -> None:
...
-                mode_key = MideaCDDevice.get_dict_key_by_value(
-                    "_modes",
-                    str(value),
+                mode_key = next(
+                    (
+                        key
+                        for key, name in self._mode_map().items()
+                        if name == str(value)
+                    ),
+                    None,
```
rokam: Same as before

midealocal/devices/cd/__init__.py
```diff
@@ -541,6 +918,22 @@ def set_attribute(self, attr: str, value: bool | float | str) -> None:
...
+                    DeviceAttributes.max_temperature_lower_limit,
+                )
+                upper = self._attributes.get(
+                    DeviceAttributes.max_temperature_upper_limit,
+                )
+                minimum = int(lower) if isinstance(lower, int | float) else 35
+                maximum = int(upper) if isinstance(upper, int | float) else 70
+                message.ts_max = max(minimum, min(maximum, int(float(value))))
+                if message.target_temperature > message.ts_max:
+                    message.target_temperature = float(message.ts_max)
```
rokam, line 937: Why set `max_temperature`? Isn't this the device limit?

rokam (changes requested):

midealocal/devices/cd/__init__.py
FrancYescO: No, on my device there is the possibility to set it in a predefined range from NetHome app

midealocal/devices/cd/__init__.py
FrancYescO: Applied in 75b8616 together with the explicit None handling. The method now uses the single combined return.

midealocal/devices/cd/__init__.py
FrancYescO: Same reason as the previous lookup: this path must accept extended-only mode names. It now uses the shared _mode_key helper added in 75b8616.

midealocal/devices/cd/__init__.py
FrancYescO: Extended protocol value 0x05 is overloaded: it normally means Heat Pump, but with the dedicated vacation flag it means Vacation. This branch returns only the canonical vacation label; it does not make the legacy mode selectable. I added an inline comment documenting the collision in 75b8616.

midealocal/devices/cd/__init__.py
FrancYescO: Addressed in 75b8616. WeeklySchedule and DailyTimerSchedule now have distinct shapes because the daily structure also contains amount, single_timer_on, single_timer_off, and timers. Validated mappings are cast only after the runtime checks pass.

midealocal/devices/cd/__init__.py
FrancYescO: The previous helper can only search a class dictionary by a fixed attribute name. Extended devices must select _modes or _extended_modes dynamically from reported capabilities. In 75b8616 I extracted the lookup into a dedicated _mode_key helper so both call sites reuse the capability-selected map without duplicating the generator.

midealocal/devices/cd/__init__.py
FrancYescO: max_temperature is the current configurable tsMax value. max_temperature_lower_limit and max_temperature_upper_limit are the immutable device bounds for that setting. The requested tsMax is clamped to those bounds, and target_temperature is lowered in the same frame only when it would otherwise exceed the new configured maximum. I added a clarifying inline comment in 75b8616.

midealocal/devices/cd/__init__.py
FrancYescO: Applied in 75b8616. The helper now accepts object | None and returns False explicitly for None.

midealocal/devices/cd/__init__.py
FrancYescO: Addressed in 75b8616 with separate TypedDict structures for weekly and daily timers. Runtime validation intentionally remains because set_attribute is an external input boundary and annotations cannot prevent malformed service payloads. The validation now also checks timer effect flags as booleans.

midealocal/devices/cd/__init__.py
rokam: Ok

rokam (approved):
LGTM

Issue #424, opened 2026-03-21
Suggestion to add missing inputs


TurboLed:
For the AC type device that I own, I have found out several missing inputs that could be intersting to add.

My device respond to group request from 0 to 15.

Groups 1,2,3,5,6,7,11,12 all contain non-zero data.

I have already submitted some of the info to the other midea-smart project:
https://github.com/mill1000/midea-msmart/issues/248
https://github.com/mill1000/midea-msmart/issues/256

I think both projects achieve the same things and should be united.

Here are some examples of inputs that seem relevant:

Group 1:
- Compressor frequency (byte 4)
- Outdoor unit current (byte 7)
- Outdoor unit voltage (byte 8)
- T1: indoor ambient temperature (byte 10 - 30) / 2
- T2: indoor coil temperature (byte 11 - 30) / 2
- T3: outdoor coil temperature (byte 12 - 50) / 2
- T4: outdoor ambient temperature (byte 13 - 50 / 2)
- TP: compressor discharge temperature (byte 14)

Group 2:
- Indoor fan speed (byte 5 * 8)

Group 5:
- outdoor fan speed (byte 8 * 8)
- expansion valve angle (byte 9 * 8)
- defrost state (byte 10)

Group 7:
- Outdoor unit power (byte 10-11)

Group 11:
- Horizontal louver angle (byte 9)
- Vertical louver angle (byte 12)

Defrost state can be useful to know when to activate auxilary heat.
T1, T2 and indoor fan speed can be useful to compute capacity (BTU/h)
Outdoor power can be useful to compute COP (capacity/power)
Horizontal angle/outdoor fan speed can be useful to know when the unit is operating.

wuwentao:
@TurboLed thank for your issue and detail info.
I have more delays these months, once time is enough, I will sync it to midea-local.
sorry.

areidyOTH:
Additional cross-source evidence for groups 3, 5, and 6 is now available.

The reference below is the latest bundled AC plugin currently present in `midea-msmart`:

https://github.com/mill1000/midea-msmart/blob/04cdb96fbcbccdf75fe0772503c3bb772ff6af89/reference/0xAC/T0xAC_3.5.260526.zip

For group 3, the plugin selects the group by the low nibble and maps these body offsets:

| Body offset | Plugin interpretation | Independent observation on MFAG70VB |
|---:|---|---|
[...]

PR #736, opened 2026-08-29
feat(ac): add group 3 outdoor fan speed

areidyOTH:
## Summary

- add the AC group 3 query
- expose decoded `outdoor_fan_speed` as read-only telemetry
- accept both `0x03` and `0x43` group 3 response discriminators
- add query, parsing, truncation, attribute-storage, and read-only tests

Related to #424.

## Protocol evidence

Tested on a Midea MFAG70VB-N / MFAG70VB-W, device type `0xAC`, protocol v3.

Group 3 request body:

`41 21 01 43 00 01 73`

An observed response body was:

`c1 21 01 03 00 00 00 00 00 08 6a 00 72 b3 92 00 16 ...`

The tested device returned response discriminator `0x03`, and body offset 10 contains `0x6a` (`106`). A bundled Midea AC plugin in `midea-msmart` selects group 3 by the low nibble, so the parser accepts both `0x03` and `0x43`.

Engineer channel 12, documented as outdoor-fan set speed, matched body offset 10 across these observations:

| Engineer display | Group 3 body offset 10 |
|---:|---:|
| 0 | 0 |
| 39 | 39 |
| 105 | 105–106 during transition |
| 106 | 106 |

The bundled plugin describes the same byte as current outdoor-fan speed and decodes it with a `raw * 8` scale. Following maintainer review, this change uses that current-speed interpretation and exposes the decoded value as `outdoor_fan_speed`. Group 5 exposes its separate target as `target_outdoor_fan_speed`.

Reference: https://github.com/mill1000/midea-msmart/blob/04cdb96fbcbccdf75fe0772503c3bb772ff6af89/reference/0xAC/T0xAC_3.5.260526.zip

## Compatibility

Group 3 uses its own query class, so devices that do not answer it can have that query excluded by the existing unsupported-protocol detection. Short responses are handled through the existing `read_byte()` default.

## Validation

- `python -m pytest ./tests/` — 1,652 passed

### DIFF
--- a/midealocal/devices/ac/__init__.py
+++ b/midealocal/devices/ac/__init__.py
@@ -18,6 +18,7 @@ from .message import (
     MessageGeneralSet,
     MessageGroupOneQuery,
     MessageGroupSevenQuery,
+    MessageGroupThreeQuery,
     MessageGroupTwoQuery,
     MessageGroupZeroQuery,
     MessageHumidityQuery,
@@ -45,6 +46,7 @@ ACQuery = (
     | MessageHumidityQuery
     | MessageGroupZeroQuery
     | MessageGroupOneQuery
+    | MessageGroupThreeQuery
     | MessageGroupTwoQuery
     | MessageGroupSevenQuery
     | MessageCapabilitiesQuery
@@ -124,6 +126,8 @@ class DeviceAttributes(StrEnum):
     indoor_fan_speed = "indoor_fan_speed"
     target_indoor_fan_speed = "target_indoor_fan_speed"
     water_pump_running = "water_pump_running"
+    # group 3: outdoor fan
+    outdoor_fan_speed_raw = "outdoor_fan_speed_raw"
     # group 7: real time compressor power
     compressor_power = "compressor_power"
 
@@ -301,6 +305,7 @@ class MideaACDevice(MideaDevice):
                 DeviceAttributes.indoor_fan_speed: None,
                 DeviceAttributes.target_indoor_fan_speed: None,
                 DeviceAttributes.water_pump_running: None,
+                DeviceAttributes.outdoor_fan_speed_raw: None,
                 DeviceAttributes.compressor_power: None,
             },
         )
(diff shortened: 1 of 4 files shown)


midealocal/devices/ac/__init__.py
```diff
@@ -124,6 +126,8 @@ class DeviceAttributes(StrEnum):
     indoor_fan_speed = "indoor_fan_speed"
     target_indoor_fan_speed = "target_indoor_fan_speed"
     water_pump_running = "water_pump_running"
+    # group 3: outdoor fan
+    outdoor_fan_speed_raw = "outdoor_fan_speed_raw"
```
rokam: Is this measured in RPM? is there another outdoor_fan_speed?
I think this could be named without _raw to be easier to understand.


```suggestion
    outdoor_fan_speed = "outdoor_fan_speed"
```

midealocal/devices/ac/message.py
```diff
@@ -1440,6 +1454,15 @@ def _parse_group_two(self, body: bytearray) -> None:
...
         self.water_pump_running = bool(body[8] & XC1_WATER_PUMP_MASK)
 
+    def _parse_group_three(self, body: bytearray) -> None:
+        """Parse group 3 data: outdoor fan speed."""
+        if len(body) < XC1_GROUP_THREE_MIN_LENGTH:
+            return
+        # Midea plugin references decode this byte as current outdoor fan speed
+        # using raw * 8. The tested unit's engineer display instead labels the
+        # matching raw byte as set speed, so preserve the unscaled value.
+        self.outdoor_fan_speed_raw = body[XC1_GROUP_THREE_SPEED_OFFSET]
```
rokam: Instead of using a guard, I think this should use the bodyparser, as bodyparsers will be a big refactor. At least change it to readByte.


```suggestion
        self.outdoor_fan_speed = self.read_byte(body, 10) * 8
```

rokam (changes requested):

midealocal/devices/ac/__init__.py
areidyOTH: Updated in 874ad4c. I renamed the attribute to `outdoor_fan_speed` and now apply the raw ×8 scaling used by the Midea plugin references for current outdoor-fan speed. Group 5 exposes the separate target value as `target_outdoor_fan_speed`.

midealocal/devices/ac/message.py
areidyOTH: Updated in 874ad4c. `_parse_group_three()` now uses `read_byte()` and applies `XC1_FAN_SPEED_FACTOR`. I retained the named `XC1_GROUP_THREE_SPEED_OFFSET` constant instead of the literal offset, following the repository’s protocol-offset convention. Short responses now default to zero through `read_byte()`, with updated tests.

rokam (approved):
Lgtm

PR #740, opened 2026-08-29
feat(ac): expose group 5 targets and defrost stage

areidyOTH:
## Summary

- extend the existing AC group 5 response parser with target outdoor-fan speed
- expose target electronic-expansion-valve angle
- expose the complete defrost stage instead of reducing it to a Boolean
- keep all three attributes read-only and preserve compatibility with short humidity-only responses

Addresses the group 5 portion of #424.

## Protocol evidence

The existing `MessageHumidityQuery` already requests group 5 (`0x45`). A bundled Midea AC plugin maps these group 5 body offsets:

| Body offset | Interpretation | Decoding |
|---:|---|---:|
| 8 | target outdoor-fan speed | `raw * 8` |
| 9 | target electronic-expansion-valve angle | `raw * 8` |
| 10 | defrost stage | integer stage |

Reference:

https://github.com/mill1000/midea-msmart/blob/04cdb96fbcbccdf75fe0772503c3bb772ff6af89/reference/0xAC/T0xAC_3.5.260526.zip

The outdoor-fan scaling also agrees with the independently implemented group 5 decoder in `midea-msmart`. That decoder currently converts offset 10 to a Boolean; retaining the integer here preserves the plugin's more precise defrost-stage semantics.

## Compatibility

- no additional query or network traffic is introduced because group 5 is already polled for indoor humidity
- each field is guarded by its own protocol offset and read with `read_byte()`
- responses long enough to contain humidity but shorter than the diagnostic fields continue to update humidity without exposing missing values
- the new attributes are initialized as unavailable and cannot generate set commands

## Validation

- `python -m pytest ./tests/` — 1,649 passed
- targeted AC tests — 149 passed

### DIFF
--- a/midealocal/devices/ac/message.py
+++ b/midealocal/devices/ac/message.py
@@ -60,11 +60,13 @@ XC1_SUBBODY_TYPE_47 = 0x47
 XC1_SUBBODY_TYPE_INDEX = 3
 XC1_HUMIDITY_INDEX = 4
 XC1_CONSUMPTION_MIN_LENGTH = 19
-XC1_GROUP_FIVE_DIAGNOSTICS_MIN_LENGTH = 11
 XC1_OPERATING_TIME_MIN_LENGTH = 19
 
 # Group data query: the third payload byte selects the group, 0x40 | group number.
 XC1_GROUP_QUERY_BASE = 0x40
+XC1_GROUP_FIVE_TARGET_OUTDOOR_FAN_SPEED_INDEX = 8
+XC1_GROUP_FIVE_TARGET_EXPANSION_VALVE_ANGLE_INDEX = 9
+XC1_GROUP_FIVE_DEFROST_STAGE_INDEX = 10
 # Minimum body length required to parse each group data response.
 XC1_GROUP_ONE_MIN_LENGTH = 15
 XC1_GROUP_TWO_MIN_LENGTH = 9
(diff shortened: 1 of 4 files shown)


midealocal/devices/ac/message.py
```diff
@@ -1440,6 +1438,18 @@ def _parse_group_two(self, body: bytearray) -> None:
...
+        """Parse group 5 data: humidity and outdoor-unit targets."""
+        if len(body) <= XC1_HUMIDITY_INDEX:
+            return
+        # Indoor humidity should match the XBB/XA1 message when available.
+        self.indoor_humidity = body[XC1_HUMIDITY_INDEX] or None
+        if len(body) < XC1_GROUP_FIVE_DIAGNOSTICS_MIN_LENGTH:
+            return
+        self.target_outdoor_fan_speed = body[8] * XC1_FAN_SPEED_FACTOR
+        self.target_expansion_valve_angle = body[9] * XC1_FAN_SPEED_FACTOR
+        self.defrost_stage = body[10]
```
rokam:
```suggestion
    def _parse_group_five(self, body: bytearray) -> None:
        """Parse group 5 data: humidity and outdoor-unit targets."""

        self.indoor_humidity = self.read_byte(body, 4)
        if(self.indoor_humidity == 0):
            self.indoor_humidity = None
        self.target_outdoor_fan_speed = self.read_byte(body, 8) * XC1_FAN_SPEED_FACTOR
        self.target_expansion_valve_angle = self.read_byte(body, 9) * XC1_FAN_SPEED_FACTOR
        self.defrost_stage = self.read_byte(body, 10)
```

rokam (changes requested):

midealocal/devices/ac/message.py
areidyOTH: Updated in 385f933. All four fields now use `read_byte()`. I retained named constants for the protocol offsets and preserved the humidity zero-to-`None` conversion using `or None`. The truncated-response test now verifies the resulting default values.

midealocal/devices/ac/message.py
```diff
@@ -1440,6 +1440,30 @@ def _parse_group_two(self, body: bytearray) -> None:
         # Could also be the float switch (tank full) that triggers the pump.
         self.water_pump_running = bool(body[8] & XC1_WATER_PUMP_MASK)
 
+    def _parse_group_five(self, body: bytearray) -> None:
+        """Parse group 5 data: humidity and outdoor-unit targets."""
+        # Indoor humidity should match the XBB/XA1 message when available.
+        if len(body) > XC1_HUMIDITY_INDEX:
```
rokam, line 1446: This kind of guard isn't needed. read_byte already checks for length

rokam (changes requested):
You need to remove the length guards, they aren't needed beacause read_byte already checks for body length.

Also, there are some conflicts to be addrassed

chemelli74:
@areidyOTH, do you have time to complete this PR ? Thx in advance

PR #744, opened 2026-08-30
refactor(message): make missing byte reads optional

areidyOTH:
## Summary

- make `MessageBody.read_byte()` return `None` when an index is absent and no fallback is supplied
- add overloads so callers that provide an integer fallback retain an `int` return type
- make every existing production parser request its current zero fallback explicitly
- add tests for present bytes, the implicit `None` fallback, explicit `None`, and custom integer fallbacks

## Motivation

The previous implicit zero fallback made an absent protocol field indistinguishable from a genuine zero reading. This is relevant to optional or short response bodies, where zero can be a valid device value.

Existing in-repository parsers keep their current runtime behavior by passing `0` explicitly. New parsers can omit the fallback when they need to preserve the distinction between an absent byte and a zero byte.

This intentionally changes the default contract for external callers that use `read_byte(body, index)` and rely on an absent index producing zero, so the PR is opened as a draft for API review.

Related discussion: #740.

## Validation

- `python -m pytest ./tests/` — 1,722 passed

### DIFF
--- a/midealocal/devices/ac/message.py
+++ b/midealocal/devices/ac/message.py
@@ -1459,7 +1459,7 @@ class XC1MessageBody(MessageBody):
         # in units of 8 RPM. The engineer display labels the matching raw byte
         # as set speed, but group 5 exposes that target separately.
         self.outdoor_fan_speed = (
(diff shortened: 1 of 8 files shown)


chemelli74:
@areidyOTH, do you have time to complete this PR ? Thx in advance

PR #754, opened 2026-09-04
feat(a1): expose device capabilities

mattrossman:
## Summary

Expose A1 device capabilities through a public capability map.

The A1 protocol already provides a `pump_enable` capability bit separately from the current `pump` state. This change exposes that existing signal through `device.capabilities["pump"]`, following the capability-reporting pattern used by other device implementations.

This supports [Home Assistant core PR #181257](https://github.com/home-assistant/core/pull/181257), which needs to create the A1 pump switch only for devices that report pump support.

## Testing

Added focused A1 device coverage verifying that the capability map reflects both supported and unsupported pump capability states reported by the device.

## Notes

Reference implementation: [midea-local PR #746](https://github.com/midea-lan/midea-local/pull/746)

### DIFF
--- a/midealocal/devices/a1/__init__.py
+++ b/midealocal/devices/a1/__init__.py
@@ -81,7 +81,7 @@ class MideaA1Device(MideaDevice):
                 DeviceAttributes.filter_cleaning_reminder: False,
             },
         )
-        self._pump_enable = False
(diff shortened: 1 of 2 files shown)


chemelli74 (approved):
LGTM, thx for the contribution !

PR #758, opened 2026-09-05
feat(b8): support the second-generation ("750004CE") protocol

chemelli74:
Adds a parallel implementation of the newer B8 vacuum protocol (cloud plugin SN8 750004CE): body-type 0xAA frames, the renumbered work-status table, task-control start/stop/pause and per-setting config messages, and the sleep/relocate/on-base sub-status. Ported from lua/b8/T_0000_B8_750004CE_2024011101.lua.

Selected per device via `_V2_SUBTYPES`, which is empty for now — v2 stays inert until a real device's cloud `subType` is added, so existing 7500001H devices are unaffected.

### DIFF
--- a/midealocal/devices/b8/message_v2.py
+++ b/midealocal/devices/b8/message_v2.py
@@ -1,0 +1,549 @@
+"""Midea local B8 second-generation ("v2") protocol messages.
+
+Newer B8 robot vacuums (Midea cloud plugin SN8 ``750004CE``, models with a
+base/wash station) speak a different wire protocol than the one in
+``message.py`` (SN8 ``7500001H``):
+
+* control/query/report frames use body-type ``0xAA`` followed by a version
+  byte ``0x01`` and a type selector, instead of body-type ``0x22``/``0x32``;
+* the work-status table is renumbered (``pause`` is ``clean_pause = 0x06``
+  here, ``0x08`` in v1, and the start/stop/pause *command* still uses the
+  legacy ``0x1B`` on its own channel);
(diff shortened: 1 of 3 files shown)


midealocal/devices/b8/message_v2.py
rokam, line 1: This V2 message should use body_parser instead of read_byte.

rokam (changes requested):

midealocal/devices/b8/message_v2.py
```diff
@@ -0,0 +1,553 @@
...
+
+    def __init__(self, body: bytearray) -> None:
+        """Parse a B8 v2 work-status body."""
+        super().__init__(
+            body,
+            [
+                IntEnumParser("work_status", _WORK_STATUS, B8V2WorkStatus),
+                IntEnumParser("control_type", _CONTROL_TYPE, B8ControlType),
+                IntEnumParser("move_direction", _MOVE_DIRECTION, B8V2Moviment),
+                IntEnumParser("clean_mode", _CLEAN_MODE, B8V2CleanMode),
```
rokam, line 461: What about describe the default value in each of those IntEnumParsers?

rokam (changes requested):

midealocal/devices/b8/message_v2.py
chemelli74: Review please

midealocal/devices/b8/message_v2.py
chemelli74: Review please

rokam (approved):
LGTM

PR #760, opened 2026-09-05
fix(cd): match controlType=0x01 SET body to lua/cd plugins

### LINKED ISSUE #468: Midea CD water heater (RSJRAC07 / 170L HPWH): WiFi `set_temperature` faults the unit — `MessageSet` body is too short (8 bytes; firmware needs 25)

chemelli74:
Every lua/cd plugin builds the controlType=0x01 SET body as at most bodyBytes[0..21] (22-byte body ending at vacationTsValue). tsMaxValue is read-only in all of them, and none write timer_type/dr_switch/schedule_mode in the SET path.

- trim MessageSet._body to the 21-byte payload; drop ts_max, _ts_max_value,
  DEFAULT_TS_MAX and schedule_mode
- make max_temperature (tsMax) and schedule_mode read-only: remove their
  controlType=0x01 write branches in set_attribute
- keep openPTC-forced-0, Tr clamp [2,6], vacation-day encoding and
  _sanitize_set_fields from #497

This replaces the speculative 25-byte body (#497/#735) that was based on an RSJRAC07 Lua not present in the repo; the repo's RSJRAC01/RSJ000CB plugins use the 22-byte body.

Closes #468

### DIFF
--- a/midealocal/devices/cd/__init__.py
+++ b/midealocal/devices/cd/__init__.py
@@ -765,38 +765,20 @@ class MideaCDDevice(MideaDevice):
             )
             return
 
-        # Power, mode, temperature, max_temperature, and vacation use controlType=0x01.
+        # Power, mode, temperature and vacation use controlType=0x01.
+        # max_temperature (tsMax) and schedule_mode are read-only: no Midea CD
+        # Lua plugin writes them in any SET body.
         if attr in [
             DeviceAttributes.mode,
             DeviceAttributes.power,
             DeviceAttributes.target_temperature,
             DeviceAttributes.vacation_mode,
             DeviceAttributes.vacation_days,
-            DeviceAttributes.max_temperature,
-            DeviceAttributes.schedule_mode,
         ]:
-            if (
-                attr
-                in {
-                    DeviceAttributes.max_temperature,
-                    DeviceAttributes.schedule_mode,
-                }
-                and not self._is_extended_water_heater()
-            ):
-                _LOGGER.warning(
-                    "[%s] %s write requires reported CD support",
-                    self.device_id,
-                    attr,
-                )
-                return
             message = MessageSet(self._message_protocol_version)
             message.fields = dict(self._fields) if self._fields else {}
             # align temperature encoding with lua protocol selection
             message.use_old_protocol = self._lua_protocol == LuaProtocol.old
-            schedule_mode = self._attributes.get(DeviceAttributes.schedule_mode)
-            message.schedule_mode = (
-                int(schedule_mode) if isinstance(schedule_mode, int | float) else 0
-            )
 
             # Get safe current values
             current_power = self._attributes.get(DeviceAttributes.power, False)
@@ -813,8 +795,7 @@ class MideaCDDevice(MideaDevice):
                 self._attributes.get(DeviceAttributes.fahrenheit, False),
             )
 
-            # full[21] vacationTsValue — not max temperature.
-            # full[23] tsMax — must be the device max (issue #468); 0 clamps SP.
+            # full[21] vacationTsValue — only sent for vacation controls.
             if attr in (
                 DeviceAttributes.target_temperature,
                 DeviceAttributes.mode,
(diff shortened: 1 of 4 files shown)


rokam (approved):
LGTM

PR #762, opened 2026-09-06
fix(ac): treat 0xFF indoor humidity as unavailable

chemelli74:


midealocal/devices/ac/message.py
```diff
@@ -1548,7 +1558,7 @@ def parse_consumption(data: bytearray) -> float:
             if len(body) <= XC1_HUMIDITY_INDEX:
                 return
             # indoor humidity, it should be the same value as XBB/XA1 message
-            self.indoor_humidity = body[4] if body[4] != 0 else None
+            self.indoor_humidity = parse_indoor_humidity(body[4])
```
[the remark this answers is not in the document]
rokam: Use the read_byte instead of the body len guard.

midealocal/devices/ac/message.py
```diff
@@ -1705,9 +1715,7 @@ def __init__(self, body: bytearray) -> None:
                         subprotocol_body[7] + subprotocol_body[8] * 256
                     ) / 100
             if subprotocol_body_len > BB_INDOOR_HUMIDITY_INDEX:
-                self.indoor_humidity = (
-                    subprotocol_body[30] if subprotocol_body[30] != 0 else None
-                )
+                self.indoor_humidity = parse_indoor_humidity(subprotocol_body[30])
```
rokam: Same, use read_byte instead of using guard.

rokam (changes requested):

midealocal/devices/ac/message.py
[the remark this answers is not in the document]
chemelli74: Done

midealocal/devices/ac/message.py
chemelli74: You suggest to use:

```python
self.indoor_humidity = parse_indoor_humidity(
                    self.read_byte(body, 30),
                )

```

correct ?

rokam: The only thing is to remove the guard and use read_byte. You can still use the parse humidity if this is used in multiple locations.

midealocal/devices/ac/message.py
```diff
@@ -1179,7 +1188,7 @@ def __init__(self, body: bytearray) -> None:
         decimal = body[18] if len(body) > TEMP_DECIMAL_MIN_BODY_LENGTH else 0
         self.indoor_temperature = self.parse_temperature(body[13], decimal & 0x0F)
         self.outdoor_temperature = self.parse_temperature(body[14], decimal >> 4)
-        self.indoor_humidity = body[17] if body[17] != 0 else None
+        self.indoor_humidity = parse_indoor_humidity(body[17])
```
rokam: Use read_byte here too.

midealocal/devices/ac/message.py
```diff
@@ -1548,7 +1558,9 @@ def parse_consumption(data: bytearray) -> float:
             if len(body) <= XC1_HUMIDITY_INDEX:
```
rokam: This can be removed when using read_byte

rokam (changes requested):

midealocal/devices/ac/message.py
```diff
@@ -1545,10 +1555,10 @@ def parse_consumption(data: bytearray) -> float:
                 + (self.current_operating_time_second / 3600)
             )
         elif group_type == XC1_SUBBODY_TYPE_45:
-            if len(body) <= XC1_HUMIDITY_INDEX:
-                return
             # indoor humidity, it should be the same value as XBB/XA1 message
-            self.indoor_humidity = body[4] if body[4] != 0 else None
+            self.indoor_humidity = parse_indoor_humidity(
+                self.read_byte(body, XC1_HUMIDITY_INDEX),
+            )
```
rokam: This logic was changed. Maybe this is the reason the tests are failing.

midealocal/devices/ac/message.py
```diff
@@ -1200,8 +1209,9 @@ def __init__(
                 params[NewProtocolTags.indirect_wind][0] == INDIRECT_WIND_VALUE
             )
         if NewProtocolTags.indoor_humidity in params:
-            indoor_humidity = params[NewProtocolTags.indoor_humidity][0]
-            self.indoor_humidity = indoor_humidity if indoor_humidity != 0 else None
+            self.indoor_humidity = parse_indoor_humidity(
+                params[NewProtocolTags.indoor_humidity][0],
+            )
```
rokam, line 1214: Logic here was changed.

midealocal/devices/ac/message.py
```diff
@@ -1704,10 +1714,9 @@ def __init__(self, body: bytearray) -> None:
                     self.indoor_temperature = (
                         subprotocol_body[7] + subprotocol_body[8] * 256
                     ) / 100
-            if subprotocol_body_len > BB_INDOOR_HUMIDITY_INDEX:
-                self.indoor_humidity = (
-                    subprotocol_body[30] if subprotocol_body[30] != 0 else None
-                )
+            self.indoor_humidity = parse_indoor_humidity(
+                self.read_byte(body, BB_INDOOR_HUMIDITY_INDEX),
```
rokam: Logic changed

rokam (changes requested):

midealocal/devices/ac/message.py
rokam: Also, the attribute is now set with none. Before it wasn't set due to early return

rokam (approved):

PR #765, opened 2026-09-08
feat: add preset modes to MideaClimateDevice

chemelli74:


midealocal/base_classes/climate.py
```diff
@@ -197,3 +204,35 @@ def set_raw_swing_mode(self, swing_mode: str) -> None:
...
+
+    @property
+    def preset_modes(self) -> Sequence[str]:
+        """Return the available preset mode names, or [] if unsupported."""
+        if not self._preset_attributes:
+            return []
+        return [PRESET_NONE, *self._preset_attributes]
+
+    @property
+    def preset_mode(self) -> str | None:
```
rokam, line 216: What do you think if we return MideaPreset strenum?

rokam (dismissed):
I think we could use strenum as presets.

midealocal/base_classes/climate.py
chemelli74: Done

rokam (dismissed):

rokam (approved):

PR #773, opened 2026-09-11
refactor(cli): cache token/key in midea-devices.json

chemelli74:
Discover now caches the token/key pair that successfully connects to a device in midea-devices.json, keyed by device_id. 

On later runs it tries that cached pair first — skipping the cloud entirely when it still works. If there's no cache yet, or the cached key no longer works (e.g. the device was re-paired), it falls back to fetching fresh credentials from the cloud as before, and re-caches whichever key succeeds.

Robustness and security hardening for the new cache:
- Malformed or unreadable cache data (bad JSON, wrong JSON shape, an entry missing token/key) is treated as a cache miss instead of crashing discover.
- The cache is written atomically (temp file + rename) so an interrupted write can't leave behind corrupt JSON.
Both midea-local.json (save) and midea-devices.json now get owner-only (0600) file permissions.
- Debug logs no longer include the raw token/key on an authentication failure.

Also adds a `--log-file` option to persist debug output to a file, and removes library_test.py, which is fully superseded by midealocal.cli discover.


rokam (dismissed):
LGTM

rokam (approved):
Lgtm

PR #375, opened 2025-04-25
feat(fa): support fa device new porotocl and args

wuwentao:
code is not ready, still in progress

PR #775, opened 2026-09-13
feat(fa): add protocol 5 support 

chemelli74:
Add protocol 5 support:

- extended modes
- voice
- scene
- target temperature / humidity

Verified against the lua reference files (T_0000_FA_560000F3_2023011001, T_0000_FA_17).

Replaces #375


rokam (approved):
LGTM

PR #781, opened 2026-09-13
fix(b3): sum remaining minutes and seconds instead of dropping one

chemelli74:


rokam (changes requested):
It looks the same for me but it's better to understand the way is written here.

Please run the tests with the previous code to check if tests passes. Passing we can merge as chore instead of fix.

rokam (approved):

PR #782, opened 2026-09-13
fix(ed): check TLV bounds before reading, not after

chemelli74:


midealocal/devices/ed/message.py
```diff
@@ -713,6 +713,10 @@ def __init__(self, body: bytearray) -> None:
         super().__init__(body)
         data_offset = 2
         while True:
+            # every branch below reads up to data_offset + 6; check that
+            # range is in bounds before touching any of it, not after.
+            if data_offset + 6 >= len(body):
+                break
             length = (body[data_offset + 2] >> 4) + 2
```
rokam:
```suggestion
            # every branch below reads up to data_offset + 6; check that
            # range is in bounds before touching any of it, not after.
            length = (body[data_offset + 2] >> 4) + 2
            if data_offset + length >= len(body):
                break
```

rokam: I think this will work better. The +6 looks like a magical number

rokam (dismissed):

midealocal/devices/ed/message.py
chemelli74: Refactored

rokam (approved):
LGTM

PR #783, opened 2026-09-13
refactor(cd): remove unused lines in sanitize_set_fields 

chemelli74:


rokam (changes requested):
This is changing this method behaviour. I think that only trValue should be returned.

midealocal/devices/cd/__init__.py
```diff
@@ -918,7 +918,7 @@ def _sanitize_set_fields(fields: dict[Any, Any]) -> dict[Any, Any]:
             tr_i = 0
         if tr_i < MessageSet.TR_VALUE_MIN or tr_i > MessageSet.TR_VALUE_MAX:
             clean.pop("trValue", None)
-        return {k: v for k, v in clean.items() if k == "trValue"}
+        return clean
```
rokam:
```suggestion
        clean = dict(fields)
        return {k: v for k, v in clean.items() if k == "trValue" and v >= MessageSet.TR_VALUE_MIN and v <= MessageSet.TR_VALUE_MAX}
```

rokam: This method can be simplified.

rokam (changes requested):

midealocal/devices/cd/__init__.py
chemelli74: Check my commit please

midealocal/devices/cd/__init__.py
```diff
@@ -903,22 +903,19 @@ def set_attribute(  # noqa: C901, PLR0911
...
+        tr = fields.get("trValue")
         try:
             tr_i = int(tr) if tr is not None else 0
         except (TypeError, ValueError):
             tr_i = 0
-        if tr_i < MessageSet.TR_VALUE_MIN or tr_i > MessageSet.TR_VALUE_MAX:
-            clean.pop("trValue", None)
-        return {k: v for k, v in clean.items() if k == "trValue"}
+        if MessageSet.TR_VALUE_MIN <= tr_i <= MessageSet.TR_VALUE_MAX:
+            return {"trValue": tr}
```
rokam:
```suggestion
            return {"trValue": tr_i}
```

rokam (changes requested):

rokam (approved):

PR #786, opened 2026-09-13
fix(ca): exception body bit offsets and add missing fields

chemelli74:


rokam (changes requested):
Those bit operations are wrong. Please check LUA.

https://github.com/midea-lan/midea-local/blob/22d92b0080cc2d6d3e46cf3ceb40287d820e69c8/lua/ca/T_0000_CA_16.lua#L948-L957

Maybe we can change this PR to fix those.

rokam (changes requested):
Those errors should be boolean instead of int.

rokam (approved):

PR #789, opened 2026-09-13
fix(b0)!: correct hot_steam mode typo

chemelli74:


rokam (approved):
This is a breaking change. The value will be different and need to be fixed inside HomeAssistant as well.

PR #801, opened 2026-09-17
fix(device): build real queries after reply with protocol version

chemelli74:
`refresh_status()` built the real status queries before the appliance-query reply resolved the device's protocol version, so they were sent with the wrong value baked in.
Devices that validate this header (see [home-assistant/core#181116](https://github.com/home-assistant/core/issues/181116)) silently dropped them and never came online.

Fix is to build the real queries only after the appliance query is sent and resolved.


rokam (approved):
LGTM

PR #807, opened 2026-09-18
fix(c3): correct UnitPara byte offsets and expose telemetry

chemelli74:
C3 byte-offset fixes (fan_speed, fg_capacity_need, current_unit_capacity, pwm_pump_out) + expose telemetry attributes


rokam (approved):
LGTM

PR #808, opened 2026-09-18
fix(e2): warn on unsupported attrs, halve read temperature

chemelli74:
Warn instead of silently dropping unsupported old-protocol attributes; halve target_temperature on read


rokam (approved):
LGTM

PR #809, opened 2026-09-18
feat(b1): decode oven mode and target_temperature

chemelli74:
Decode oven mode and target_temperature from the X01 body


rokam (approved):
LGTM

PR #810, opened 2026-09-18
fix(ac): drop error_code_query default, derive rate_select gears

chemelli74:
Drop error_code_query from default query list; derive rate_select gears from capability


rokam (approved):
LGTM

